PR #2, opened 2020-08-28
deploy etcd from subchart

kd7lxl:
This adds support for installing etcd via the bitnami chart and configures apiserver to connect to it.

To get the secret names to match up between etcd and the apiserver deployment, an override is needed. I've included a makefile to assist in setting that. You can run `make deploy-konk` to deploy an instance of konk to your default cluster and namespace. Works with minikube.

⚠️ This requires users to change their CRs, which super-sucks:
https://github.com/infobloxopen/konk/blob/5ee6f3e87c6ffefb0033cf6e125b85f4cab3e464/examples/konk.yaml#L1-L9


charts/konk/Chart.lock
```diff
@@ -0,0 +1,6 @@
+dependencies:
+- name: etcd
+  repository: https://charts.bitnami.com/bitnami
+  version: 4.10.1
+digest: sha256:ff4979a73febb97d7f588ef06e58070e0458a6454aa66749e2bd1d1d9cd93798
+generated: "2020-08-27T15:59:52.948412-07:00"
```
drewwells, line 6: We should use the Etcd bundled in Kind. It's tested to work with the apiserver. The entrypoint starts all of the services using systemd https://github.com/kubernetes-sigs/kind/blob/master/images/base/Dockerfile#L127

charts/konk/templates/etcd-secret.yaml
```diff
@@ -0,0 +1,17 @@
...
+apiVersion: v1
+kind: Secret
+type: Opaque
+metadata:
+  name: {{ include "konk.fullname" . }}-etcd-cert
+  labels:
+    {{- include "konk.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": "pre-install"
+    "helm.sh/hook-delete-policy": "before-hook-creation"
```
drewwells, line 11: hooks aren't required. K8s will prevent the pod from initializing until this secret is created. If we do use them, please add a comment that delete-policy is mandatory when using pre-install hook

kd7lxl:
Alternative etcd provisioning in #4

drewwells:
We have etcd in the chart now. Is this still needed?

kd7lxl:
> We have etcd in the chart now. Is this still needed?

This is an alternative way of provisioning etcd that can provide HA when configured to do so. That is the only potential use of this I can think of.

drewwells:
@kd7lxl rebase the pr

helm-charts/konk/values.yaml
```diff
@@ -63,6 +63,22 @@ etcd:
...
     # runAsUser: 1000
 
+  # subchart options
+  subchart: true
+  auth:
+    rbac:
+      enabled: false
+    client:
+      secureTransport: true
+      existingSecret: # must override this with RELEASE_NAME-CHART_NAME-etcd-cert
```
kd7lxl, line 73: I'm not sure how to deal with this in the operator. We shouldn't have to leave it up to the user to define in their CR, since it could be constructed from the name of the CR.

kd7lxl: We might be able to flip it around: let the etcd chart generate the certs, and point to that secret in the apiserver deployment.

examples/konk.yaml
```diff
@@ -2,4 +2,8 @@ apiVersion: konk.infoblox.com/v1alpha1
 kind: Konk
 metadata:
   name: example-konk
-spec: {}
+spec:
+  etcd:
+    auth:
+      client:
+        existingSecret: example-konk-etcd-cert
```
drewwells, line 9: Hard to use this generally since the CR creator needs to know the etcd secret name

examples/konk.yaml
```diff
@@ -2,4 +2,8 @@ apiVersion: konk.infoblox.com/v1alpha1
 kind: Konk
 metadata:
   name: runner-konk
-spec: {}
+spec:
+  etcd:
+    auth:
+      client:
+        existingSecret: runner-konk-etcd-cert
```
drewwells, line 9: try adding bitnami etcd chart to konk-operator

kd7lxl:
Duplicate of #135

PR #10, opened 2020-09-03
Added code to create and deploy certs and kubeconfig

abalaven:
Summary of changes- 

Apiserver and Etcd certs and kubeconfig are now created while deploying the konk chart - 

```
sc-ml-avenkata:konk abalavenkata$ make deploy-konk
helm  upgrade -i abalavenkata-konk charts/konk
Release "abalavenkata-konk" does not exist. Installing it now.
NAME: abalavenkata-konk
LAST DEPLOYED: Thu Sep  3 16:21:51 2020
NAMESPACE: default
STATUS: deployed
REVISION: 1
NOTES:
1. Get the application URL by running these commands:
  export POD_NAME=$(kubectl get pods --namespace default -l "app.kubernetes.io/name=konk,app.kubernetes.io/instance=abalavenkata-konk" -o jsonpath="{.items[0].metadata.name}")
  echo "Visit http://127.0.0.1:8080 to use your application"
  kubectl --namespace default port-forward $POD_NAME 8080:80
sc-ml-avenkata:konk abalavenkata$ kubectl get pods
NAME                                 READY   STATUS      RESTARTS   AGE
abalavenkata-konk-f57446765-wtwk2    2/2     Running     0          21s
abalavenkata-konk-kind-pod           0/1     Completed   0          21s
sc-ml-avenkata:konk abalavenkata$ kubectl get secret
NAME                                      TYPE                                  DATA   AGE
abalavenkata-konk-etcd-cert               Opaque                                5      7m57s <- CERTS
abalavenkata-konk-kubeconfig              Opaque                                1      7m56s <- KUBECONFIG
abalavenkata-konk-token-ckv5n             kubernetes.io/service-account-token   3      27s
default-token-8gzxk                       kubernetes.io/service-account-token   3      147d
sh.helm.release.v1.abalavenkata-konk.v1   helm.sh/release.v1                    1      27s
```


charts/konk/templates/configmap.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "konk.fullname" . }}-configmap
+  labels:
+    app: {{ include "konk.name" . }}
+    chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
```
kd7lxl, line 9: Use the `include` for this to keep it consistent with the others.

charts/konk/templates/configmap.yaml
```diff
@@ -0,0 +1,23 @@
...
+    VAL=$(kubectl -n {{ .Release.Namespace }} get secret {{ include "konk.fullname" . }}-etcd-cert --ignore-not-found)
+    if [[ $VAL == "" ]]
+    then
+      kubectl -n {{ .Release.Namespace }} create secret generic {{ include "konk.fullname" . }}-etcd-cert --from-file=/etc/kubernetes/pki/etcd/ca.crt --from-file=/etc/kubernetes/pki/etcd/server.crt --from-file=/etc/kubernetes/pki/etcd/server.key --from-file=/etc/kubernetes/pki/apiserver-etcd-client.crt --from-file=/etc/kubernetes/pki/apiserver-etcd-client.key
+    fi
+    VAL=$(kubectl -n {{ .Release.Namespace }} get secret {{ include "konk.fullname" . }}-kubeconfig --ignore-not-found)
+    if [[ $VAL == "" ]]
+    then
+      kubectl -n {{ .Release.Namespace }} create secret generic {{ include "konk.fullname" . }}-kubeconfig --from-file=/etc/kubernetes/admin.conf
+    fi
```
kd7lxl, line 23: Please move the bash code into charts/konk/files/start.sh (or maybe provision.sh) and load it into the ConfigMap like this: https://helm.sh/docs/chart_template_guide/accessing_files/#configmap-and-secrets-utility-functions. This way we can use the bash script outside of the helm chart if we want, like for local development.

charts/konk/templates/configmap.yaml
```diff
@@ -0,0 +1,23 @@
...
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+data:
+  start.sh: |
+    kubeadm init phase certs all
+    kubeadm init phase kubeconfig admin
+    VAL=$(kubectl -n {{ .Release.Namespace }} get secret {{ include "konk.fullname" . }}-etcd-cert --ignore-not-found)
+    if [[ $VAL == "" ]]
+    then
+      kubectl -n {{ .Release.Namespace }} create secret generic {{ include "konk.fullname" . }}-etcd-cert --from-file=/etc/kubernetes/pki/etcd/ca.crt --from-file=/etc/kubernetes/pki/etcd/server.crt --from-file=/etc/kubernetes/pki/etcd/server.key --from-file=/etc/kubernetes/pki/apiserver-etcd-client.crt --from-file=/etc/kubernetes/pki/apiserver-etcd-client.key
```
kd7lxl, line 17: It would be good to put these into two different secrets, one for the apiserver container and one for the etcd container, for the concept of "least privilege". The containers don't need access to each others private keys.

charts/konk/templates/kind-pod.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: {{ include "konk.fullname" . }}-kind-pod
```
kd7lxl, line 4: There's got to be a better name for this... maybe init?
```suggestion
  name: {{ include "konk.fullname" . }}-init
```

charts/konk/templates/kind-pod.yaml
```diff
@@ -0,0 +1,34 @@
...
+        containerPort: 443
+        protocol: TCP
+    resources:
+      {{- toYaml .Values.kind.resources | nindent 6 }}
+    volumeMounts:
+    - mountPath: /kind/
+      name: certs-configmap
+      readOnly: true
+  volumes:
+  - name: certs-configmap
```
kd7lxl, line 30: The naming here is a little misleading. I'd probably call it `scripts`
```suggestion
    volumeMounts:
    - mountPath: /scripts/
      name: scripts
      readOnly: true
  volumes:
  - name: scripts
```

charts/konk/templates/configmap.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "konk.fullname" . }}-configmap
```
kd7lxl, line 4:
```suggestion
  name: {{ include "konk.fullname" . }}-scripts
```

charts/konk/templates/role.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
```
kd7lxl, line 3: Add labels.

charts/konk/templates/rolebinding.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
```
kd7lxl, line 3: Add labels.

charts/konk/templates/kind-pod.yaml
```diff
@@ -0,0 +1,34 @@
...
+  name: {{ include "konk.fullname" . }}-kind-pod
+  labels:
+    {{- include "konk.labels" . | nindent 4 }}
+spec:
+  serviceAccountName: {{ include "konk.serviceAccountName" . }}
+  containers:
+  - name: kind
+    securityContext:
+      {{- toYaml .Values.kind.securityContext | nindent 6 }}
+    image: "{{ .Values.kind.image.repository }}:{{ .Values.kind.image.tag }}"
```
kd7lxl, line 13: I would default this to the appVersion like your comment describes.

charts/konk/templates/kind-pod.yaml
```diff
@@ -0,0 +1,34 @@
...
+    imagePullPolicy: {{ .Values.kind.image.pullPolicy }}
+    command:
+      - bash
+    args:
+      - /kind/start.sh
+    ports:
+      - name: https
+        containerPort: 443
+        protocol: TCP
```
kd7lxl, line 22: Does this pod actually expose a webserver? Probably shouldn't define a port if it's not a service.

charts/konk/templates/configmap.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
```
kd7lxl, line 3: Oh, we haven't done this with any of the others yet, but we need to start adding `metadata.namespace` to all of the resources so they land in the right namespace even when Spinnaker doesn't override the namespace.

charts/konk/templates/configmap.yaml
abalaven: Updated the labels.

charts/konk/templates/configmap.yaml
abalaven: Moved the bash script to provision.sh under the scripts directory.

charts/konk/templates/configmap.yaml
abalaven: Split the certs into 2 separate secrets for apiserver and etcd.

charts/konk/templates/kind-pod.yaml
abalaven: Changed as per suggestion.

charts/konk/templates/kind-pod.yaml
abalaven: Done

charts/konk/templates/configmap.yaml
abalaven: Done

charts/konk/templates/role.yaml
abalaven: Done

charts/konk/templates/rolebinding.yaml
abalaven: Done

charts/konk/templates/kind-pod.yaml
abalaven: Updated to default to the appVersion.

charts/konk/templates/kind-pod.yaml
abalaven: Removed the port config

charts/konk/templates/configmap.yaml
abalaven: Added Namespace info to all the resources.

kd7lxl (approved):
Looks good, thanks.

charts/konk/scripts/provision.sh
```diff
@@ -0,0 +1,17 @@
+kubeadm init phase certs all
+kubeadm init phase kubeconfig admin
+VAL=$(kubectl -n {{ .Release.Namespace }} get secret {{ include "konk.fullname" . }}-etcd-cert --ignore-not-found)
+if [[ $VAL == "" ]]
```
kd7lxl, line 4: Instead of string matching, you can just use the exit code.
```suggestion
if ! kubectl -n {{ .Release.Namespace }} get secret {{ include "konk.fullname" . }}-etcd-cert
```

kd7lxl: Speaking of which, we should probably improve the script to error out when `kubectl` requests fail.

charts/konk/scripts/provision.sh
```diff
@@ -0,0 +1,17 @@
+kubeadm init phase certs all
```
drewwells, line 1: Exit on errors and spit out every command that is executed. Use /bin/sh if bash is not available in the pod
```suggestion
#!/bin/bash -xe

kubeadm init phase certs all
```

charts/konk/scripts/provision.sh
drewwells: If you take off `--ignore-not-found` exclude `-e` from the setopts above

drewwells (approved):

charts/konk/scripts/provision.sh
kd7lxl: No, you still want the `-e`. Since it's in an `if`, it won't cause the whole script to fail. Demo:
```
thayward@tac-ml-thayward ~ % cat test.sh                                                                                                                                       env-4
set -xe
if ! kubectl get ns foobar
then
	echo true
fi
thayward@tac-ml-thayward ~ % ./test.sh                                                                                                                                         env-4
+ kubectl get ns foobar
Error from server (NotFound): namespaces "foobar" not found
+ echo true
true
```

kd7lxl (approved):

charts/konk/scripts/provision.sh
abalaven: Updated to use exit code instead of string matching.

charts/konk/scripts/provision.sh
abalaven: Done

PR #11, opened 2020-09-04
Add test for apiserver with kubeconfig

kd7lxl:
This adds a helm test pod that connects to apiserver using the generated kubeconfig and runs a sample command we always expect to succeed, `kubectl get ns`. This serves as both an integration test and as an example for konk users. It is exercised by Github Actions CI to prove konk works.


abalaven (approved):

.github/workflows/pr.yaml
```diff
@@ -14,4 +14,13 @@ jobs:
...
         run: |
-          make helm-lint HELM="docker run --rm -i -v $PWD:/apps infoblox/helm:3.2.4-5b243a2"
+          mkdir -p ~/.kube
+          make helm-lint
+      - uses: helm/kind-action@v1.0.0
+      - name: Test
+        working-directory: ${{ github.workspace }}/src/github.com/${{ github.repository }}
+        run: |
+          kubectl cluster-info
+          make deploy-konk
```
drewwells, line 24: Why isn't this using --wait?

charts/konk/templates/tests/test-connection.yaml
```diff
@@ -8,8 +8,27 @@ metadata:
...
-      image: busybox
-      command: ['wget']
-      args: ['{{ include "konk.fullname" . }}:{{ .Values.service.port }}']
+  - name: kind
+    image: "{{ .Values.kind.image.repository }}:{{ .Values.kind.image.tag | default .Chart.AppVersion }}"
+    imagePullPolicy: {{ .Values.kind.image.pullPolicy }}
+    command:
+      - kubectl
+    args:
+      - --server=https://{{ include "konk.fullname" . }}:{{ .Values.service.port }}
```
drewwells, line 17: Todo default kubeconfig to fqdn service name

charts/konk/templates/deployment.yaml
```diff
@@ -37,19 +38,37 @@ spec:
...
+              path: /healthz
+              port: https
+              scheme: HTTPS
+          {{- if .Values.apiserver.startupProbe }}
+          startupProbe:
+            httpGet:
+              path: /healthz
+              port: https
+              scheme: HTTPS
+          {{- end }}
```
drewwells, line 63: Neat!

.github/workflows/pr.yaml
kd7lxl: The static init pod never becomes ready, so `--wait` blocks forever. We can add `--wait` once this issue is resolved.

drewwells (approved):

charts/konk/templates/tests/test-connection.yaml
kd7lxl: Completed in https://github.com/infobloxopen/konk/pull/16

PR #12, opened 2020-09-08
support --wait during install

kd7lxl:
https://github.com/infobloxopen/konk/pull/11#discussion_r483896814 requested adding `--wait`. Since helm does not consider Completed pods ready, we can support `--wait` by making the init script hang instead of completing.

### DIFF
--- a/charts/konk/scripts/provision.sh
+++ b/charts/konk/scripts/provision.sh
@@ -22,3 +22,5 @@ then
   kubectl -n {{ .Release.Namespace }} create secret generic {{ include "konk.fullname" . }}-kubeconfig \
     --from-file=/etc/kubernetes/admin.conf
 fi
+
+sleep 300
--- a/charts/konk/scripts/provision.sh
+++ b/charts/konk/scripts/provision.sh
@@ -23,4 +23,7 @@ then
     --from-file=/etc/kubernetes/admin.conf
 fi
 
+echo
+date
+echo done.
 sleep 300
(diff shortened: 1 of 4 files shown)


charts/konk/templates/init-pod.yaml
```diff
@@ -29,3 +29,4 @@ spec:
       name: {{ include "konk.fullname" . }}-scripts
       defaultMode: 0777
   restartPolicy: OnFailure
+  terminationGracePeriodSeconds: 10
```
abalaven, line 32: Do we need to add this?

kd7lxl: Need? Not really, but it speeds up termination of the pod. The default is 30 seconds, and the script doesn't react to SIGTERM, so we always have to wait for the termination grace period for the pod to terminate.

abalaven (approved):

charts/konk/scripts/provision.sh
```diff
@@ -22,3 +22,8 @@ then
   kubectl -n {{ .Release.Namespace }} create secret generic {{ include "konk.fullname" . }}-kubeconfig \
     --from-file=/etc/kubernetes/admin.conf
 fi
+
+echo
+date
+echo done.
+sleep 300
```
kd7lxl, line 29: This only works for the first 300 seconds, after which a `helm upgrade -i --wait` will hang forever due to the Completed pod. I'll experiment with other solutions.

PR #15, opened 2020-09-09
disable unsupported APIs

kd7lxl:
KONK runs an apiserver independently without a node and CNI, so many features of Kubernetes like Deployments are unsupported. This PR disables them so that users will get an error when trying to create these types of resources.


drewwells (approved):
seems like a good idea anyways

PR #16, opened 2020-09-09
generate certs and kubeconfig with correct service name

kd7lxl:
The kubeconfig now trusts the apiserver.

Resolves TODO: https://github.com/infobloxopen/konk/pull/11#discussion_r483897211

### DIFF
--- a/charts/konk/scripts/provision.sh
+++ b/charts/konk/scripts/provision.sh
@@ -1,7 +1,9 @@
-#!/bin/bash -xe
+#!/bin/bash
+set -xe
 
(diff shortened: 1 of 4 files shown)


charts/konk/scripts/provision.sh
```diff
@@ -1,7 +1,9 @@
-#!/bin/bash -xe
```
kd7lxl, line 1: This inline config was not actually setting the bash options when run with `helm test`.

drewwells (approved):

PR #23, opened 2020-09-12
Feature/helm2

drewwells:
adds a rule to generate CRD from kustomize file
ran make bundle hoping it would regenerate crd, included the output

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ helm-lint: helm-lint-$(notdir $(CHART_DIR)/*)
 helm-lint-%:
 	$(HELM) lint $(CHART_DIR)/$*
 
-%-konk-operator: HELM_FLAGS += --set=image.tag=$(GIT_VERSION)
+%-konk-operator: HELM_FLAGS += --set=image.tag=$(GIT_VERSION) --crds.create=true
 
 deploy-%:
 	$(HELM) upgrade -i $(RELEASE_NAME)-$* $(CHART_DIR)/$* $(HELM_FLAGS)
@@ -132,3 +132,8 @@ bundle: kustomize
 .PHONY: bundle-build
 bundle-build:
 	docker build -f bundle.Dockerfile -t $(BUNDLE_IMG) .
+
+# TODO Replace with controller_gen
+manifests: $(KUSTOMIZE)
+	$(KUSTOMIZE) build config/crd/ > .tmp.konk
+	mv .tmp.konk config/crd/bases/konk.infoblox.com_konks.yaml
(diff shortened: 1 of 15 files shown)


Makefile
```diff
@@ -132,3 +132,7 @@ bundle: kustomize
 .PHONY: bundle-build
 bundle-build:
 	docker build -f bundle.Dockerfile -t $(BUNDLE_IMG) .
+
+# TODO Replace with controller_gen
+manifests: $(KUSTOMIZE)
+	$(KUSTOMIZE) build config/crd > config/crd/bases/konk.infoblox.com_konks.yaml
```
kd7lxl, line 138: Does this replace the kustomize template with the kustomize rendered yaml?

config/crd/bases/konk.infoblox.com_konks.yaml
```diff
@@ -1,44 +0,0 @@
----
-apiVersion: apiextensions.k8s.io/v1
-kind: CustomResourceDefinition
```
kd7lxl, line 3: The CRD is empty/removed now?

helm-charts/konk-operator/templates/crds.yaml
```diff
@@ -0,0 +1,6 @@
+{{- if .Values.crds.create }}
+{{- range $path, $bytes := .Files.Glob "crds/*.yaml" }}
```
kd7lxl, line 2: `crds/*.yaml` is symlinked to `config/crd/bases`, which is empty now, so I don't think this will work.

Makefile
drewwells: the configuration is in config/crd/kustomization.yaml. config/crd/bases/*.yaml are created from that config

kd7lxl: Maybe this should write to `helm-charts/konk-operator/crds/`

drewwells: hmm it deleted the file now... wut

Makefile
```diff
@@ -132,3 +132,8 @@ bundle: kustomize
 .PHONY: bundle-build
 bundle-build:
 	docker build -f bundle.Dockerfile -t $(BUNDLE_IMG) .
+
+# TODO Replace with controller_gen
+manifests: $(KUSTOMIZE)
+	$(KUSTOMIZE) build config/crd/ > .tmp.konk
+	mv .tmp.konk config/crd/bases/konk.infoblox.com_konks.yaml
```
drewwells, line 139: kustomize uses config/crd/bases/* as input... can't write back to it directly

helm-charts/konk-operator/templates/crds.yaml
drewwells: looks like we can't symlink, package is also failing

kd7lxl (approved):

kd7lxl (approved):

PR #25, opened 2020-09-14
runnable readme

drewwells:


kd7lxl (approved):
Maybe one day we can use this trick to inject malicious code into a user's cluster.

PR #29, opened 2020-09-18
move init to a deployment so that it is mutable

kd7lxl:
This moves the static init pod into a deployment as an init pod. The deployment then runs `sleep infinity` as its primary container command.

#### Pros:
* provisioning script is mutable / can be upgraded
* `helm upgrade --wait` works

#### Cons:
* Requires a persistent pod that just runs `sleep` all the time. It does not request any resources, so at least it is not reserving any quota even if it is consuming a small bit of memory to stay running.

I tried using helm hooks to manage the static pod (recreate it on upgrade) or as a job, but this doesn't really work. If the hook is `pre-install`, the RBAC, ServiceAccount, and ConfigMap used by the init pod don't exist and it fails. When configured as a `post-install` hook, the job doesn't run until after all deployments become ready when using `--wait`, which doesn't work for konk which needs the init pod to create secrets before the deployment becomes ready.


helm-charts/konk/templates/init.yaml
```diff
@@ -0,0 +1,61 @@
...
+      {{- include "konk.selectorLabels" . | nindent 6 }}
+      app.kubernetes.io/component: init
+  template:
+    metadata:
+      labels:
+        {{- include "konk.selectorLabels" . | nindent 8 }}
+        app.kubernetes.io/component: init
+    spec:
+      serviceAccountName: {{ include "konk.serviceAccountName" . }}
+      initContainers:
```
abalaven, line 21: Just to clarify, since this is a separate deployment, the `initContainer` approach will work since the secrets being created here are mounted to the pod created by the other deployment. Correct?

helm-charts/konk/templates/init.yaml
```diff
@@ -0,0 +1,61 @@
...
+          readOnly: true
+      containers:
+      - name: done
+        securityContext:
+          {{- toYaml .Values.kind.securityContext | nindent 10 }}
+        image: "{{ .Values.kind.image.repository }}:{{ .Values.kind.image.tag | default .Chart.AppVersion }}"
+        imagePullPolicy: {{ .Values.kind.image.pullPolicy }}
+        command:
+          - sleep
+          - infinity
```
abalaven, line 56: Shouldn't infinity be an `arg`?

kd7lxl: Maybe it could be, but this is working as-is.

helm-charts/konk/templates/init.yaml
kd7lxl: Yes, this deployment/pod has no dependency on the secrets it creates. The other deployment does have a dependency on these secrets, which is why the provision.sh cannot be in that other deployment.

abalaven (approved):

PR #32, opened 2020-09-25
Create Issuer and Certificate resources for allowing public authenticated traffic

abalaven:
Added code to create an Issuer and Certificate (Cert-manager CRDs), which creates a secret to be referenced to by service ingress object.

Generates the following objects - 

 Issuer - 

```
konk abalavenkata(k8s: minikube) $ kubectl get issuer -n ab -o yaml
apiVersion: v1
items:
- apiVersion: cert-manager.io/v1
  kind: Issuer
  metadata:
    annotations:
      meta.helm.sh/release-name: abalavenkata-konk
      meta.helm.sh/release-namespace: ab
    creationTimestamp: "2020-09-25T04:13:24Z"
    generation: 1
    labels:
      app.kubernetes.io/managed-by: Helm
    name: abalavenkata-konk-ca-issuer
    namespace: ab
    resourceVersion: "2374043"
    selfLink: /apis/cert-manager.io/v1/namespaces/ab/issuers/abalavenkata-konk-ca-issuer
    uid: d895ff41-4850-427f-bc93-03e04e5faff4
  spec:
    ca:
      secretName: abalavenkata-konk-ca
  status:
    conditions:
    - lastTransitionTime: "2020-09-25T04:14:18Z"
      message: Signing CA verified
      reason: KeyPairVerified
      status: "True"
      type: Ready
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```

Certificate -

```
konk abalavenkata(k8s: minikube) $ kubectl get certificate -n ab -o yaml
apiVersion: v1
items:
- apiVersion: cert-manager.io/v1
  kind: Certificate
  metadata:
    annotations:
      meta.helm.sh/release-name: abalavenkata-konk
      meta.helm.sh/release-namespace: ab
    creationTimestamp: "2020-09-25T04:13:24Z"
    generation: 1
    labels:
      app.kubernetes.io/managed-by: Helm
    name: abalavenkata-konk-certificate
    namespace: ab
    resourceVersion: "2374068"
    selfLink: /apis/cert-manager.io/v1/namespaces/ab/certificates/abalavenkata-konk-certificate
    uid: 05b1eb7a-7c96-46e3-af41-7963728c2b07
  spec:
    dnsNames:
    - abalavenkata-konk.ab.svc.cluster.local
    issuerRef:
      kind: Issuer
      name: abalavenkata-konk-ca-issuer
    secretName: abalavenkata-konk-generated-cert
  status:
    conditions:
    - lastTransitionTime: "2020-09-25T04:14:20Z"
      message: Certificate is up to date and has not expired
      reason: Ready
      status: "True"
      type: Ready
    notAfter: "2020-12-24T04:14:19Z"
    notBefore: "2020-09-25T04:14:19Z"
    renewalTime: "2020-11-24T04:14:19Z"
    revision: 1
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```

Generated Secret - 

```
konk abalavenkata(k8s: minikube) $ kubectl get secret -n ab
NAME                                      TYPE                                  DATA   AGE
abalavenkata-konk-apiserver-cert          Opaque                                6      26m
abalavenkata-konk-ca                      kubernetes.io/tls                     2      26m
abalavenkata-konk-etcd-cert               Opaque                                3      26m
abalavenkata-konk-generated-cert          kubernetes.io/tls                     3      25m <<<<<< GENERATED CERT
abalavenkata-konk-kubeconfig              Opaque                                1      26m
abalavenkata-konk-token-wlxgk             kubernetes.io/service-account-token   3      26m
default-token-mdz57                       kubernetes.io/service-account-token   3      21d
sh.helm.release.v1.abalavenkata-konk.v1   helm.sh/release.v1                    1      26m

```
**!TEST PENDING!**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ helm-lint-%:
 
 deploy-cert-manager:
 	kubectl create namespace cert-manager
-	$(HELM) install --wait cert-manager --namespace cert-manager jetstack/cert-manager --version v1.0.1 \
+	$(HELM) upgrade -i --wait cert-manager --namespace cert-manager jetstack/cert-manager --version v1.0.1 \
 		--set installCRDs=true \
 		--set extraArgs[0]="--enable-certificate-owner-ref=true"
 
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ helm-lint-%:
 
 deploy-cert-manager:
 	kubectl create namespace cert-manager
-	$(HELM) install cert-manager --namespace cert-manager jetstack/cert-manager --version v1.0.1 \
+	$(HELM) install --wait cert-manager --namespace cert-manager jetstack/cert-manager --version v1.0.1 \
 		--set installCRDs=true \
 		--set extraArgs[0]="--enable-certificate-owner-ref=true"
 
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,13 @@ helm-lint: helm-lint-$(notdir $(CHART_DIR)/*)
 helm-lint-%:
 	$(HELM) lint $(CHART_DIR)/$*
 
+deploy-cert-manager:
+	kubectl create namespace cert-manager
+	$(HELM) repo add jetstack https://charts.jetstack.io
+	$(HELM) install cert-manager --namespace cert-manager jetstack/cert-manager --version v1.0.1 \
+		--set installCRDs=true \
+		--set extraArgs[0]="--enable-certificate-owner-ref=true"
+
 %-konk-operator: HELM_FLAGS ?= --set=image.tag=$(GIT_VERSION) --set=image.pullPolicy=IfNotPresent
 
 deploy-%: package
--- a/Makefile
+++ b/Makefile
@@ -30,8 +30,8 @@ helm-lint-%:
 	$(HELM) lint $(CHART_DIR)/$*
 
 deploy-cert-manager:
-	kubectl create namespace cert-manager
 	$(HELM) upgrade -i --wait cert-manager --namespace cert-manager jetstack/cert-manager --version v1.0.1 \
+		--create-namespace \
 		--set installCRDs=true \
 		--set extraArgs[0]="--enable-certificate-owner-ref=true"
 
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,8 @@ HELM		?= docker run --rm -i \
 			-e KUBECONFIG=/apps/.kube/$(notdir $(KUBECONFIG)) \
 			-v $(dir $(KUBECONFIG)):/apps/.kube/ \
 			-v $(PWD):/apps \
-			-u $(shell id -u):$(shell id -g) \
+			-v $(PWD)/repositories.yaml:/tmp/.config/helm/repositories.yaml \
+			-v $(PWD)/jetstack-index.yaml:/tmp/.cache/helm/repository/jetstack-index.yaml \
 			infoblox/helm:3.2.4-5b243a2 \
 			helm
 K8S_RELEASE	?= v1.19.0
@@ -30,7 +31,6 @@ helm-lint-%:
 
 deploy-cert-manager:
 	kubectl create namespace cert-manager
-	$(HELM) repo add jetstack https://charts.jetstack.io
 	$(HELM) install cert-manager --namespace cert-manager jetstack/cert-manager --version v1.0.1 \
 		--set installCRDs=true \
 		--set extraArgs[0]="--enable-certificate-owner-ref=true"
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -68,7 +68,9 @@ rules:
   - "roles"
   - "rolebindings"
 - verbs:
-  - "*"
+  - get
+  - create
+  - list
   apiGroups:
   - "cert-manager.io"
   resources:
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -67,5 +67,12 @@ rules:
   resources:
   - "roles"
   - "rolebindings"
+- verbs:
+  - "*"
+  apiGroups:
+  - "cert-manager.io"
+  resources:
+  - "issuers"
+  - "certificates"
 
 # +kubebuilder:scaffold:rules
--- a/helm-charts/konk/templates/certificate.yaml
+++ b/helm-charts/konk/templates/certificate.yaml
@@ -1,0 +1,12 @@
+apiVersion: cert-manager.io/v1alpha2
+kind: Certificate
+metadata:
+  name: {{ include "konk.fullname" . }}-certificate
+  namespace: {{ .Release.Namespace }}
+spec:
+  secretName: {{ include "konk.fullname" . }}-generated-cert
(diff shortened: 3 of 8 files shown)


drewwells:
There's a fake repositories.yaml file in that container. See if removing it helps with that helm repo add error https://github.com/infobloxopen/helm/blob/master/Dockerfile.in#L13

abalaven:
I tried that but it didn't work. Still seeing the same error.
I also tried to deploy the cert-manager chart without the `helm repo add`, but got another error - 

`Error: mkdir /tmp/.cache/helm/repository: permission denied`

helm-charts/konk/templates/certificate.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: cert-manager.io/v1alpha2
+kind: Certificate
+metadata:
+  name: {{ include "konk.fullname" . }}-certificate
+  namespace: {{ .Release.Namespace }}
```
kd7lxl, line 5: Include the labels like in the other resources: https://github.com/infobloxopen/konk/blob/cf35713a340b9967d56f179319da2b10a6af6570/helm-charts/konk/templates/configmap.yaml#L6-L7

helm-charts/konk/templates/issuer.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: cert-manager.io/v1alpha2
+kind: Issuer
+metadata:
+  name: {{ include "konk.fullname" . }}-ca-issuer
+  namespace: {{ .Release.Namespace }}
```
kd7lxl, line 5: Include the labels like in the other resources: https://github.com/infobloxopen/konk/blob/cf35713a340b9967d56f179319da2b10a6af6570/helm-charts/konk/templates/configmap.yaml#L6-L7

helm-charts/konk/templates/certificate.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: cert-manager.io/v1alpha2
+kind: Certificate
+metadata:
+  name: {{ include "konk.fullname" . }}-certificate
```
kd7lxl, line 4: Can we use a name that describes its purpose a little better? Is this cert used for ingress support? Maybe this would be a better name:
```suggestion
  name: {{ include "konk.fullname" . }}-ingress-client
```

jetstack-index.yaml
```diff
@@ -0,0 +1,1098 @@
+apiVersion: v1
```
kd7lxl, line 1: This file should be generated/pulled on demand rather than committed to the repo.

Makefile
```diff
@@ -28,6 +29,12 @@ helm-lint: helm-lint-$(notdir $(CHART_DIR)/*)
 helm-lint-%:
 	$(HELM) lint $(CHART_DIR)/$*
 
+deploy-cert-manager:
+	kubectl create namespace cert-manager
+	$(HELM) install --wait cert-manager --namespace cert-manager jetstack/cert-manager --version v1.0.1 \
```
kd7lxl, line 34: Another way to install this would be as an optional dependency of helm-operator. What do you prefer @drewwells?

jetstack-index.yaml
abalaven: I agree. However, this file and `repositories.yaml` are supposed to be generated when we do a `helm repo add`. But since every helm command is run by a separate container, we need to conserve the state after `helm repo add`. Is it possible to have a single container execute both `helm repo add` and `helm install cert-manager`?

Makefile
drewwells: This is guaranteed to fail on every run after the initial install. I'd keep it out of our chart and it needs to check if cert manager is already in the cluster. If you ran this on env-4, it's very likely to break certs in the cluster. In authz we check if we're in minikube before doing crd installs

drewwells: https://github.com/Infoblox-CTO/ngp.authz/blob/7ca636f56270a77fe4566336bf77deb278611469/Makefile

abalaven: My understanding is that the `deploy-cert-manager` target will only be used for testing in this repo. Since cert-manager already exists on env-4, and other envs by extension, we don't need to call this target at all. Please correct me if I am wrong.

drewwells: Yeah that's fine. I commented from my phone and thought it was being referenced in a more generic deploy rule.

Makefile
```diff
@@ -28,6 +29,12 @@ helm-lint: helm-lint-$(notdir $(CHART_DIR)/*)
 helm-lint-%:
 	$(HELM) lint $(CHART_DIR)/$*
 
+deploy-cert-manager:
+	kubectl create namespace cert-manager
```
drewwells, line 33: helm will create the namespace
```suggestion
```

Makefile
drewwells, line 34:
```suggestion
	$(HELM) upgrade -i --wait cert-manager --namespace cert-manager jetstack/cert-manager --version v1.0.1 \
```
This will do upgrades and installs

config/rbac/role.yaml
```diff
@@ -67,5 +67,12 @@ rules:
   resources:
   - "roles"
   - "rolebindings"
+- verbs:
+  - "*"
```
drewwells, line 71: I believe we can get by with just crate here
```suggestion
  - "CREATE"
```

Makefile
abalaven: Doesn't look like it. I got this error when I removed the create namespace step - 

```
docker run --rm -i --entrypoint="" --network host -e KUBECONFIG=/apps/.kube/config -v /Users/abalavenkata/.kube/:/apps/.kube/ -v /Users/abalavenkata/go/src/github.com/abalaven/konk:/apps -v /Users/abalavenkata/go/src/github.com/abalaven/konk/repositories.yaml:/tmp/.config/helm/repositories.yaml -v /Users/abalavenkata/go/src/github.com/abalaven/konk/jetstack-index.yaml:/tmp/.cache/helm/repository/jetstack-index.yaml infoblox/helm:3.2.4-5b243a2 helm upgrade -i --wait cert-manager --namespace cert-manager jetstack/cert-manager --version v1.0.1 \
		--set installCRDs=true \
		--set extraArgs[0]="--enable-certificate-owner-ref=true"
Release "cert-manager" does not exist. Installing it now.
Error: create: failed to create: namespaces "cert-manager" not found
make: *** [deploy-cert-manager] Error 1
```

Makefile
abalaven: Done

config/rbac/role.yaml
abalaven: Looks like we need Create, Get and List. Without these, the konk operator does not deploy a konk instance.

Makefile
drewwells: They changed the default behavior in helm3, you have to pass a flag to create the namespace https://helm.sh/docs/faq/#automatically-creating-namespaces

drewwells (approved):

Makefile
abalaven: Done, updated to include the `--create-namespace` flag.

PR #34, opened 2020-10-05
add make rules for standing up kind

drewwells:
Add some of the github-actions steps to the makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,13 @@ KUBEADM		?= docker run --rm -it --entrypoint="" kindest/node:$(K8S_RELEASE) kube
 KUBECONFIG	?= ${HOME}/.kube/config
 RELEASE_NAME	?= $(USER)
 
+# KIND env variables
+KIND_NAME   	?= konk
+NODE_VERSION    ?= v1.19.0
+NODE_IMAGE      ?= kindest/node:${NODE_VERSION}
+KIND_VERSION    ?= v0.8.1
+KIND 			:= bin/kind
+
 default: all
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -35,10 +42,10 @@ deploy-cert-manager:
 		--set installCRDs=true \
 		--set extraArgs[0]="--enable-certificate-owner-ref=true"
 
-%-konk-operator: HELM_FLAGS ?= --set=image.tag=$(GIT_VERSION) --set=image.pullPolicy=IfNotPresent
+%-konk-operator: HELM_FLAGS ?= --set=image.pullPolicy=IfNotPresent
 
 deploy-%: package
-	$(HELM) upgrade -i --wait $(RELEASE_NAME)-$* $(CHART_DIR)/$* $(HELM_FLAGS)
+	$(HELM) upgrade -i --wait $(RELEASE_NAME)-$* $(CHART_DIR)/$* --set=image.tag=$(GIT_VERSION) $(HELM_FLAGS)
```
kd7lxl, line 48: It's sloppy to set `image.tag` on `make deploy-konk`, even if that value is currently unused in that chart. That is why the flag previously only applied to `*-konk-operator`.

drewwells: not sure I am following. konk and konk-operator would both have the same version from git

kd7lxl: konk's image tag does not track the git version; it tracks k8s/apiserver/etcd versions. So you don't want to be overriding `image.tag` on konk with the git version.

PR #35, opened 2020-10-05
Generate helm repo add files on demand

abalaven:
Added an extra target in Makefile - 

`HELM_CM` - Performs `helm repo add` and `helm install cert-manager`. Deploys cert-manager with the files generated by `helm repo add`.


Makefile
```diff
@@ -1,15 +1,30 @@
...
-			-v $(PWD)/repositories.yaml:/tmp/.config/helm/repositories.yaml \
-			-v $(PWD)/jetstack-index.yaml:/tmp/.cache/helm/repository/jetstack-index.yaml \
 			infoblox/helm:3.2.4-5b243a2 \
 			helm
+HELM_RA         ?= docker run -i \
+                        --entrypoint="" \
+                        --network host \
+			--name $(DOCKER_NAME) \
+                        infoblox/helm:3.2.4-5b243a2 \
+                        helm
```
kd7lxl, line 17: Since most of this is in common for all three commands, you could use a variable to define this part once and then construct each command by using the variable.

abalaven: Done

Makefile
```diff
@@ -30,10 +42,16 @@ helm-lint-%:
 	$(HELM) lint $(CHART_DIR)/$*
 
 deploy-cert-manager:
-	$(HELM) upgrade -i --wait cert-manager --namespace cert-manager jetstack/cert-manager --version v1.0.1 \
+	$(HELM_RA) repo add jetstack https://charts.jetstack.io
+	docker cp $(DOCKER_NAME):/tmp/.config/helm/repositories.yaml .
+	docker cp $(DOCKER_NAME):/tmp/.cache/helm/repository/jetstack-index.yaml .
+	docker rm $(DOCKER_NAME)
+	$(HELM_CM) upgrade -i --wait cert-manager --namespace cert-manager jetstack/cert-manager --version v1.0.1 \
```
kd7lxl, line 49: Have you considered combining this into a single command, something like `docker run ... sh -c "helm repo add ... && helm upgrade ...`? Then you could skip the copying of files in and out of the container.

kd7lxl (approved):

Makefile
abalaven: That is definitely a lot cleaner. Made the change.

Makefile
```diff
@@ -1,15 +1,19 @@
 CHART_DIR	:= helm-charts
 GIT_VERSION	?= $(shell git describe --dirty=-unsupported --always --long --tags)
+HELM_IMAGE	?= infoblox/helm:3.2.4-5b243a2
+DOCKER_FLAGS	?= --entrypoint="" \
+                        --network host \
+                        -e KUBECONFIG=/apps/.kube/$(notdir $(KUBECONFIG)) \
+                        -v $(dir $(KUBECONFIG)):/apps/.kube/ \
+                        -v $(PWD):/apps
```
kd7lxl, line 8: You could more in the `DOCKER_FLAGS`...
```suggestion
DOCKER_RUNNER	?= docker run --rm -i \
                        --entrypoint="" \
                        --network host \
                        -e KUBECONFIG=/apps/.kube/$(notdir $(KUBECONFIG)) \
                        -v $(dir $(KUBECONFIG)):/apps/.kube/ \
                        -v $(PWD):/apps
                        $(HELM_IMAGE)
```

Makefile
```diff
@@ -1,15 +1,17 @@
...
-			-v $(PWD)/repositories.yaml:/tmp/.config/helm/repositories.yaml \
-			-v $(PWD)/jetstack-index.yaml:/tmp/.cache/helm/repository/jetstack-index.yaml \
-			infoblox/helm:3.2.4-5b243a2 \
+HELM_IMAGE	?= infoblox/helm:3.2.4-5b243a2
+DOCKER_RUNNER	?= docker run --rm -i \
+                        --entrypoint="" \
+                        --network host \
+                        -e KUBECONFIG=/apps/.kube/$(notdir $(KUBECONFIG)) \
+                        -v $(dir $(KUBECONFIG)):/apps/.kube/ \
+                        -v $(PWD):/apps \
```
kd7lxl, line 9: Can you restore the spaces to tabs here? The rest of the file has tabs, and most of these lines don't even need to change in this PR.

Makefile
```diff
@@ -29,11 +31,12 @@ helm-lint: helm-lint-$(notdir $(CHART_DIR)/*)
...
 	$(HELM) lint $(CHART_DIR)/$*
 
+# Run this only if your cluster does not have cert-manager already deployed
 deploy-cert-manager:
-	$(HELM) upgrade -i --wait cert-manager --namespace cert-manager jetstack/cert-manager --version v1.0.1 \
+	$(HELM_CMD) "helm repo add jetstack https://charts.jetstack.io && helm upgrade -i --wait cert-manager --namespace cert-manager jetstack/cert-manager --version v1.0.1 \
 		--create-namespace \
 		--set installCRDs=true \
-		--set extraArgs[0]="--enable-certificate-owner-ref=true"
+		--set extraArgs[0]="--enable-certificate-owner-ref=true""
```
kd7lxl, line 39: Does the double quote here actually work?

kd7lxl (approved):

Makefile
abalaven: Tested manually after each change and the Github Check is working too. 😄

PR #38, opened 2020-10-07
build binary independent of apiserver-boot

drewwells:


test/apiserver/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM infobloxcto/alpine_base:0.1.0 AS runner
```
drewwells, line 1: This repo is public now. We could move it to the opensource account..

kd7lxl (approved):

PR #40, opened 2020-10-19
Apiserver chart

drewwells:
DEMO

```
apiVersion: apiregistration.k8s.io/v1
kind: APIService
metadata:
  annotations:
    meta.helm.sh/release-name: drew-example-apiserver
    meta.helm.sh/release-namespace: default
  creationTimestamp: "2020-10-20T04:23:43Z"
  labels:
    api: contact
    apiserver: "true"
    app.kubernetes.io/managed-by: Helm
  name: v1alpha1.example.infoblox.com
  resourceVersion: "119394"
  selfLink: /apis/apiregistration.k8s.io/v1/apiservices/v1alpha1.example.infoblox.com
  uid: 7bf90e38-c957-4c1e-89e5-9a34c6d4aecf
spec:
...
status:
  conditions:
  - lastTransitionTime: "2020-10-20T04:24:12Z"
    message: all checks passed
    reason: Passed
    status: "True"
    type: Available
```


drewwells:
It still has some issues, but reports ready now
```
k get contact                                                                 kind-minikube
Error from server (NotFound): Unable to list "example.infoblox.com/v1alpha1, Resource=contacts": the server could not find the requested resource
```

.github/workflows/pr.yaml
```diff
@@ -1,25 +1,38 @@
 name: PR
 on: pull_request
 jobs:
-  build:
-    name: Build
+  end2end:
+    name: End To End
```
drewwells, line 5: We can separate out building go code and pushing docker images from the helm deployments eventually. There's not really much code to do this with though, so the renaming may have been a preoptimization.

kd7lxl (approved):

PR #44, opened 2020-10-23
create service in konk

kd7lxl:
Demo: APIServer is available
```
% k exec -it example-konkservice-konk-service-kubectl-6b7df94766-82px4 -- kubectl get apiservice v1alpha1.example.infoblox.com
NAME                            SERVICE                            AVAILABLE   AGE
v1alpha1.example.infoblox.com   default/runner-example-apiserver   True        32m
```

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -61,8 +61,8 @@ jobs:
             sleep 1
           done
           POD_NAME=$(kubectl get pod -o name | grep konkservice)
-          kubectl exec -it $POD_NAME -- kubectl describe apiservice v1alpha1.runner-example-apiserver.infoblox.com
-          kubectl exec -it $POD_NAME -- kubectl get apiservice v1alpha1.runner-example-apiserver.infoblox.com | grep True
+          kubectl exec -it $POD_NAME -- kubectl describe apiservice v1alpha1.example.infoblox.com
+          kubectl exec -it $POD_NAME -- kubectl get apiservice v1alpha1.example.infoblox.com | grep True
           kubectl exec -it $POD_NAME -- kubectl get crd appconfigs.stable.example.com websites.extensions.example.com
       - name: Print failure
         timeout-minutes: 1
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -61,8 +61,11 @@ jobs:
             sleep 1
           done
           POD_NAME=$(kubectl get pod -o name | grep konkservice)
-          kubectl exec -it $POD_NAME -- kubectl describe apiservice v1alpha1.example.infoblox.com
-          kubectl exec -it $POD_NAME -- kubectl get apiservice v1alpha1.example.infoblox.com | grep True
+          until kubectl exec -it $POD_NAME -- kubectl get apiservice v1alpha1.example.infoblox.com | grep True
+          do
+            kubectl exec -it $POD_NAME -- kubectl describe apiservice v1alpha1.example.infoblox.com
+            sleep 10
+          done
           kubectl exec -it $POD_NAME -- kubectl get crd appconfigs.stable.example.com websites.extensions.example.com
       - name: Print failure
         timeout-minutes: 1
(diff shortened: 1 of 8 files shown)


abalaven (approved):

.github/workflows/pr.yaml
```diff
@@ -58,16 +66,9 @@ jobs:
             sleep 1
           done
           POD_NAME=$(kubectl get pod -o name | grep konkservice)
-          kubectl exec -it $POD_NAME -- kubectl get apiservice v1alpha1.test.infoblox.com
+          kubectl exec -it $POD_NAME -- kubectl describe apiservice v1alpha1.test.infoblox.com
+          kubectl exec -it $POD_NAME -- kubectl get apiservice v1alpha1.test.infoblox.com | grep True
```
abalaven, line 70: If we add this, the check on github will fail consistently, since we are not currently testing with a service as apart of the e2e test. (Unless there is plans to add one subsequently).

kd7lxl: Aren't we? apiserver gets deployed during the test.

abalaven: Oh that's right, we re-ordered that.

helm-charts/example-konk-service/templates/konk-service.yaml
```diff
@@ -6,6 +6,6 @@ spec:
   konk:
     name: example-konk
   service:
-    name: test
+    name: {{ .Release.Name }}-example-apiserver
```
abalaven, line 9: I think you can change this to `runner-example-apiserver`.

kd7lxl: `runner-example-apiserver` is only valid in Github Actions, where the username is `runner`. I was hoping for a name that would work everywhere, but this isn't quite it. Locally, this turns into `thayward-example-konk-service-example-apiserver`, which isn't right either.

abalaven: This value should be overridden when the KonkService CR is used in any service chart, so I think for this example CR, we can just set a static `service.name` as `runner-example-apiserver`, since the Github action is currently the only one who uses this example CR.

I'm making some changes in the README to include info about the new charts which will elaborate more on the actual values required.

kd7lxl: I changed it, but also changed the makefile to always call it `runner-example-apiserver` so that it works everywhere.

helm-charts/konk-service/templates/configmap.yaml
```diff
@@ -25,18 +22,32 @@ data:
...
+      namespace: {{` {{ NAMESPACE }} `}}
+    spec:
+      type: ExternalName
+      externalName: {{` {{ SERVICENAME }} `}}.{{` {{ NAMESPACE }} `}}
+    ---
     apiVersion: apiregistration.k8s.io/v1
     kind: APIService
     metadata:
-      name: v1alpha1.{{` {{ SERVICENAME }} `}}.infoblox.com
+      name: {{ .Values.version }}.{{ .Values.group }}
```
abalaven, line 42: This worked? 😮 Interesting.

kd7lxl: This part of the template is evaluated by the helm-operator at the time it deploys the chart from the CR. So as long as the value is set in the CR, it will get populated here. It wouldn't work for values generated at runtime like the cert contents.

abalaven: Oh, this is just the name, nvm.

.github/workflows/pr.yaml
```diff
@@ -44,10 +44,14 @@ jobs:
...
           done
-      - name: Deploy konk-service
+      - name: Setup Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.15.x
+      - name: Deploy Test APIServer
         timeout-minutes: 5
         run: |
-          make deploy-konk-service
```
kd7lxl, line 50: Since this is deployed later by the operator, it's a waste of time to deploy it here too.

abalaven: True, I included it because we deploy both the konk chart and konk CR (along with the operator), but since we are not running any test on the konk-service chart, it may not be necessary.

abalaven (approved):
LGTM, thanks

PR #45, opened 2020-10-23
remove konk-service-operator

kd7lxl:
konk-operator handles both konk and konkservice CRs, so there's no need for konk-service-operator


.github/workflows/pr.yaml
```diff
@@ -48,10 +48,9 @@ jobs:
         timeout-minutes: 5
         run: |
           make deploy-konk-service
-      - name: Deploy konk-service-operator
+      - name: Deploy konk-service
```
abalaven, line 51: Can you make this `Deploy example-konk-service`, so that it is distinct from the previous name?

Makefile
```diff
@@ -139,15 +137,7 @@ konk-operator-${GIT_VERSION}.tgz:
...
-konk-service-operator-${GIT_VERSION}.tgz:
-	mkdir -p helm-charts/konk-service-operator/crds
-	cp -vR config/crd/bases/* helm-charts/konk-service-operator/crds/
-	cp -vR config/rbac helm-charts/konk-service-operator/
-	rm helm-charts/konk-service-operator/rbac/konk_editor_role.yaml
-	rm helm-charts/konk-service-operator/rbac/konk_viewer_role.yaml
-	${HELM} package helm-charts/konk-service-operator --version ${GIT_VERSION} --app-version ${GIT_VERSION}
-
-package: konk-operator-${GIT_VERSION}.tgz konk-${GIT_VERSION}.tgz konk-service-operator-${GIT_VERSION}.tgz konk-service-${GIT_VERSION}.tgz
+package: konk-operator-${GIT_VERSION}.tgz konk-${GIT_VERSION}.tgz konk-service-${GIT_VERSION}.tgz
```
abalaven, line 140: Looks good, but the rm `konkservice_editor_role.yaml` and `rm konkservice_viewer_role.yaml` need to be removed from  the `konk-operator-${GIT_VERSION}.tgz` target

abalaven (approved):

PR #48, opened 2020-10-25
TK-65403 disable custom rest backend

drewwells:
TK-65403 disable custom rest backend

    - custom rest does not register any resource verbs kubernetes-sigs/apiserver-builder-alpha#551
    - template etcd variables
    - use temp dir by default
    - test example apiserver in PR build

DEMO
```
k apply -f sample/contact.yaml                                  
contact.example.infoblox.com/contact-example created
k get contact                                                   
NAME              CREATED AT
contact-example   2020-10-25T03:35:23Z
```


kd7lxl (approved):

kd7lxl (commented):
You could run your demo around this point of the e2e test:
https://github.com/infobloxopen/konk/blob/a2abc8a269f3a3af9fc76b5834a0b63eef5c6041/.github/workflows/pr.yaml#L69

abalaven (approved):

drewwells:
Added a PR step that verified contact CR can be created. The helm chart exists --wait without the API Service actually being up. This is disappointing. The until retries will ensure actions can be performed.

https://github.com/infobloxopen/konk/pull/48/checks

helm-charts/example-apiserver/values.yaml
```diff
@@ -97,7 +97,7 @@ tolerations: []
 affinity: {}
 
 apiserver:
-  create: false
+  create: true
```
kd7lxl, line 100: This creates the APIServer outside of konk, so what's the point? We want to prove it works inside konk.

PR #50, opened 2020-10-26
Don't put -konk in kubeconfig name

kd7lxl:
By using the exact release name (not a template) to build the kubeconfig name, it is easier for other charts that need to reference that kubeconfig to predict the secret name. Before this, the logic of the fullname template would sometimes insert `-konk` into the name and sometimes not, making it difficult to predict (dependent charts do not have access to konk's fullname template use it to generate the secret name directly).


helm-charts/konk-service/templates/configmap.yaml
```diff
@@ -13,7 +13,16 @@ data:
...
+    delay=2
+    until kubectl apply -f /gen/api-service.yaml
+    do
+      sleep $delay
+      let delay="delay*2"
+      if [ "$delay" -gt "60" ]
+      then
+        delay=60
+      fi
+    done
```
kd7lxl, line 25: This retries adding the APIService to konk until konk is ready. Before adding this, it sometimes failed when racing to create the APIService before konk was ready.

abalaven (approved):

PR #51, opened 2020-10-26
Define default resources for konk

kd7lxl:
Ideally, all users of the Konk CR would define right-size resources for konk's containers in their Konk spec, but I don't realistically expect this to happen. Reasonable default values, provided in this PR, will be much better than nothing for most cases, and can still be overridden in their CR as needed.

The apiserver and etcd requests in this PR are based on current measurements in a dev cluster. The init container resources are a guess because I couldn't figure out how to measure it (its lifetime is too short for kube-state-metrics to catch it).

An entirely alternative approach could be to use defaulting in the CRD to provide these defaults.
https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting

### DIFF
--- a/helm-charts/konk/values.yaml
+++ b/helm-charts/konk/values.yaml
@@ -10,17 +10,13 @@ apiserver:
     pullPolicy: Always
     # Overrides the image tag whose default is the chart appVersion.
     tag: ""
-  resources: {}
-    # We usually recommend not to specify default resources and to leave this as a conscious
-    # choice for the user. This also increases chances charts run on environments with little
-    # resources, such as Minikube. If you do want to specify resources, uncomment the following
-    # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
-    # limits:
-    #   cpu: 100m
-    #   memory: 128Mi
-    # requests:
-    #   cpu: 100m
-    #   memory: 128Mi
+  resources:
+    limits:
+      cpu: 200m
+      memory: 512Mi
+    requests:
+      cpu: 20m
+      memory: 160Mi
   securityContext: {}
     # capabilities:
     #   drop:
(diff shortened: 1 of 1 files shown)


drewwells (approved):
I'm surprised it works on minikube, but apparently minikube now inherits the host cpu/memory as its node capacity. Maybe default resources aren't such a bad thing now 🥳 

I still think this belongs in deployment-configurations/envs/konk-values.yaml instead. 

```
Addresses:
  InternalIP:  172.17.0.3
  Hostname:    minikube-worker
Capacity:
 cpu:                12
 ephemeral-storage:  1922728840Ki
 hugepages-1Gi:      0
 hugepages-2Mi:      0
 memory:             32805392Ki
 pods:               110
```

kd7lxl:
> I still think this belongs in deployment-configurations/envs/konk-values.yaml instead.

There is no such thing, because konks are created by the operator. And I don't see a way to provide default values for a specific CR in a helm operator so that it would apply to all konks deployed by an operator. We can bake the defaults into the CRD, but then they'll be global rather than environment specific.

abalaven (approved):

drewwells:
It's done through overrideValues https://sdk.operatorframework.io/docs/building-operators/helm/reference/advanced_features/override_values/

drewwells:
I think this is fine for now

kd7lxl:
> It's done through overrideValues https://sdk.operatorframework.io/docs/building-operators/helm/reference/advanced_features/override_values/

It says:
> By setting `image.repository` to `quay.io/mycustomrepo` you are ensuring that `quay.io/mycustomrepo` will always be used instead of the chart’s default repository (`nginx`). If the CR attempts to set this value, it will be ignored.

This is not the behavior I want. I want to specify a default that the CR can override. I don't see an existing feature that provides that behavior.

PR #52, opened 2020-10-26
Require konk and service name fields in KonkService

kd7lxl:
This updates the KonkService CRD to require the konk.name and service.name fields.


abalaven (approved):
Can we add `version` and `group` here as well? Those will be required values too.

kd7lxl:
> Can we add `version` and `group` here as well? Those will be required values too.

Makes sense 👍

abalaven (approved):

PR #55, opened 2020-10-26
Combine example-apiserver and example-konk-service charts

abalaven:
Moved the `KonkService` CR and example CRDs to the `example-apiserver` chart. The `KonkService` CR is now created when we deploy the `example-apiserver` chart.

Also updated README to elaborate on the new charts added to the repo.


helm-charts/example-apiserver/templates/konk-service.yaml
```diff
@@ -6,8 +6,8 @@ spec:
   konk:
     name: example-konk
```
kd7lxl, line 7: As a more complete example, it would be good to set this with values.

helm-charts/example-apiserver/templates/konk-service.yaml
```diff
@@ -6,8 +6,8 @@ spec:
...
   service:
-    name: runner-example-apiserver
-  group: example.infoblox.com
-  version: v1alpha1
+    name: {{ include "example-apiserver.fullname" . }}
+  group: {{ .Values.group }}
+  version: {{ .Values.version }}
```
kd7lxl, line 11: This will likely be static for any given app, so they don't really need to be values.

abalaven: I figured it would be easier to add these fields to values, since its also used here (forgot to override)- https://github.com/infobloxopen/konk/blob/main/helm-charts/example-apiserver/templates/apiservice.yaml#L18-L19

Although I don't believe we need to create this in the external cluster.

.github/workflows/pr.yaml
```diff
@@ -52,6 +52,17 @@ jobs:
         timeout-minutes: 5
         run: |
           make kind-load-apiserver deploy-example-apiserver KIND_NAME="chart-testing"
+          until kubectl wait --timeout=2m --for=condition=ready pod -l app.kubernetes.io/component=apiservice,app.kubernetes.io/instance=example-konkservice
+          do
+            sleep 1
+          done
```
drewwells, line 58: You can remove this, the helm command uses `--wait` which implements a more complete version of this ready pod check here.

.github/workflows/pr.yaml
```diff
@@ -52,6 +52,17 @@ jobs:
...
+          POD_NAME=$(kubectl get pod -o name | grep konkservice)
+          until kubectl exec -it $POD_NAME -- kubectl get apiservice v1alpha1.example.infoblox.com | grep True
+          do
+            kubectl exec -it $POD_NAME -- kubectl describe apiservice v1alpha1.example.infoblox.com
+            sleep 10
+          done
+          kubectl exec -it $POD_NAME -- kubectl get crd appconfigs.stable.example.com websites.extensions.example.com
```
drewwells, line 65: Can you move the apiservice check to here too? Then, Test APIServer actions block can be removed
```suggestion
          until kubectl apply -f test/apiserver/sample/contact.yaml
          do
            sleep 5
          done
          until kubectl get contact
          do
            sleep 5
          done
          kubectl exec -it $POD_NAME -- kubectl get crd appconfigs.stable.example.com websites.extensions.example.com          
```

drewwells (commented):
The get pod check should be removed. helm upgrade --wait does a better job of checking readiness and is in the make rule

kd7lxl (approved):

helm-charts/konk-service/values.yaml
```diff
@@ -48,3 +48,9 @@ service:
   name: test
 
 crds: null
+
+group:
+
+version:
+
+tlsVerify: false
```
kd7lxl, line 56: We should default this to `true`.

kd7lxl: You're really confusing me now. Is this feature verifying or skipping verification? If the variable is named opposite of what it does, the name should be changed.

abalaven: You're right. Will rename it to `tlsVerifySkip`.

kd7lxl: It would be least confusing to name it exactly after the k8s field: `insecureSkipTLSVerify`
https://github.com/infobloxopen/konk/blob/4ac9771d63fcad89759c33758111042a5244b962/helm-charts/konk-service/templates/configmap.yaml#L59

helm-charts/konk-service/templates/configmap.yaml
```diff
@@ -50,9 +50,14 @@ data:
...
       name: {{ .Values.version }}.{{ .Values.group }}
     spec:
+      {{- if eq .Values.insecureSkipTLSVerify false }}
       caBundle: {{` {{ CERT }} `}}
+      {{- end }}
       group: {{ .Values.group }}
       groupPriorityMinimum: 1000
+      {{- if eq .Values.insecureSkipTLSVerify true }}
+      insecureSkipTLSVerify: {{ .Values.insecureSkipTLSVerify }}
+      {{- end }}
```
kd7lxl, line 60: Now that you're not trying to match null values as true, this logic could be simplified:
```suggestion
      {{- if .Values.insecureSkipTLSVerify }}
      insecureSkipTLSVerify: {{ .Values.insecureSkipTLSVerify }}
      {{- else }}
      caBundle: {{` {{ CERT }} `}}
      {{- end }}
      group: {{ .Values.group }}
      groupPriorityMinimum: 1000
```

kd7lxl (approved):

helm-charts/konk-service/templates/configmap.yaml
abalaven: Done

PR #57, opened 2020-10-28
TK-65894 add rbac aggregation to admin and edit roles

drewwells:
TK-65894 add rbac aggregation to admin and edit roles

should be merged after #55

### DIFF
--- a/examples/konk-service.yaml
+++ b/examples/konk-service.yaml
@@ -3,7 +3,17 @@ kind: KonkService
 metadata:
   name: example-konkservice
 spec:
-  group: example.infoblox.com
+  group:
+    name: example.infoblox.com
+    kinds:
+    - Contact
+    verbs:
+    - create
+    - update
+    - get
+    - list
+    - delete
+    - watch
   konk:
     name: example-konk
   service:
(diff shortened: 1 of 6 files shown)


helm-charts/konk-service/values.yaml
```diff
@@ -50,6 +50,18 @@ service:
 crds: null
 
 group:
+  name: example.com
```
abalaven, line 53: It looks like this is relevant only to the `example-apiserver` chart, can we add this directly to the `KonkService` object in that chart?

examples/konk-service.yaml
```diff
@@ -3,7 +3,17 @@ kind: KonkService
 metadata:
   name: example-konkservice
 spec:
-  group: example.infoblox.com
+  group:
+    name: example.infoblox.com
+  kinds:
```
abalaven, line 8: `kinds` and `verbs` should be indented the same way as `name`

helm-charts/konk-service/values.yaml
drewwells: I changed the parameters. Group is now an object with properties so we can configure rbac permissions

abalaven (approved):

PR #58, opened 2020-10-28
drews computer needs longer to start kube-apiserver

drewwells:


helm-charts/konk/templates/deployment.yaml
```diff
@@ -67,6 +67,8 @@ spec:
               path: /healthz
               port: https
               scheme: HTTPS
+            failureThreshold: 30
```
kd7lxl, line 70: `30 * 10s = 5m` to finish startup. Seems fine.

kd7lxl (approved):

helm-charts/konk/templates/deployment.yaml
drewwells: I skimmed the docs while doing this and found this in an example. I don't think this impacts the liveness probe that was causing me issues

PR #59, opened 2020-10-29
use konk service in pr build

drewwells:


.github/workflows/pr.yaml
```diff
@@ -27,31 +28,30 @@ jobs:
...
-          make test-konk
       - name: Test konk-operator
         timeout-minutes: 5
         run: |
-          make kind-load-konk KIND_NAME="chart-testing"
-          make deploy-konk-operator
+          make -e kind-load-konk-operator
+          make -e deploy-konk-operator
           kubectl create -f examples/konk.yaml
           until kubectl wait --timeout=3m --for=condition=ready pod -l app.kubernetes.io/component=apiserver,app.kubernetes.io/instance=example-konk
```
kd7lxl, line 39: It's named `example` now.
```suggestion
          until kubectl wait --timeout=3m --for=condition=ready pod -l app.kubernetes.io/component=apiserver,app.kubernetes.io/instance=example
```

kd7lxl: Take one down, pass it around, test is still failing.

PR #62, opened 2020-10-29
add support for x-remote-headers

drewwells:


pjferrell (approved):
lgtm

helm-charts/konk/values.yaml
```diff
@@ -25,6 +25,11 @@ apiserver:
     # runAsNonRoot: true
     # runAsUser: 1000
   startupProbe: true
+  remoteHeaders:
+    - --requestheader-extra-headers-prefix=X-Remote-Extra-
+    - --requestheader-group-headers=X-Remote-Group
+    - --requestheader-username-headers=X-Remote-User
+  extraFlags: []
```
kd7lxl, line 32: If this were a map it could be additive, which might be easier to deal with in a system of layered values overrides.

kd7lxl (approved):

helm-charts/konk/values.yaml
drewwells: I agree

drewwells:
DEMO
```
k get po -o yaml example-konk-5959d66ddb-4g69x | grep requestheader
    - --requestheader-extra-headers-prefix=X-Remote-Extra-
    - --requestheader-group-headers=X-Remote-Group
    - --requestheader-username-headers=X-Remote-User
```

PR #66, opened 2020-10-30
init container manager

drewwells:
Block manager until apiservice is responding. This will prevent the manager crashloops

### DIFF
--- a/helm-charts/example-apiserver/templates/konk-service.yaml
+++ b/helm-charts/example-apiserver/templates/konk-service.yaml
@@ -1,7 +1,7 @@
 apiVersion: konk.infoblox.com/v1alpha1
 kind: KonkService
 metadata:
-  name: {{ include "example-apiserver.fullname" . }}-api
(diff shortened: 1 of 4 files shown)


helm-charts/example-apiserver/templates/konk-service.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: konk.infoblox.com/v1alpha1
 kind: KonkService
 metadata:
-  name: {{ include "example-apiserver.fullname" . }}-api
+  name: {{ .Release.Name }}-api
```
drewwells, line 4: The pod names were getting pretty long. kubectl pod is now named `drew-apiserver-api-konk-service-kubectl-5b6dcc84b6-2tkmt`

drewwells: -api is required b/c the release names have to be unique across all charts managed by helm-operator(s)

PR #86, opened 2020-11-02
deploy KONK as part of example-apiserver

drewwells:
- put back example used in documentation
- add helm test to example-apiserver

TODO:
- make example-apiserver chart work with a pre-existing KONK


test/apiserver/makefile
```diff
@@ -40,11 +43,21 @@ config: config/konk.yaml config/infra.yaml
...
+
+kind-load-etcd: .kind-load-etcd-${ETCD_TAG}
+
+.kind-load-${IMAGE_TAG}: .image-${IMAGE_TAG}
+	touch $@
+
+kind-load-apiserver: .kind-load-${IMAGE_TAG}
+
+# does nothing unless a KIND cli is provided
+kind-load: KIND=echo
```
kd7lxl, line 59: `echo`?

drewwells: easy way to make this work in a non-kind cluster

test/apiserver/makefile
```diff
@@ -40,11 +43,21 @@ config: config/konk.yaml config/infra.yaml
...
 	kubectl apply -f $<
 
-kind-load: KIND=false
-kind-load: QUAY_IMG=quay.io/coreos/etcd:v3.4.13
-kind-load: image
-	docker pull ${QUAY_IMG}
-	$(KIND) load docker-image ${QUAY_IMG} --name ${KIND_NAME}
+.kind-load-etcd-${ETCD_TAG}:
+	docker pull ${ETCD_IMG}
+	$(KIND) load docker-image ${ETCD_IMG} --name ${KIND_NAME}
```
kd7lxl, line 48: Why not just allow kind to pull the image from quay.io?

test/apiserver/makefile
```diff
@@ -40,11 +43,21 @@ config: config/konk.yaml config/infra.yaml
...
+	touch $@
+
+kind-load-etcd: .kind-load-etcd-${ETCD_TAG}
+
+.kind-load-${IMAGE_TAG}: .image-${IMAGE_TAG}
+	touch $@
```
kd7lxl, line 54: The lifecycle of this file doesn't match the lifecycle of the kind cluster. This will lead to inconsistency.

.github/workflows/pr.yaml
```diff
@@ -59,16 +59,13 @@ jobs:
...
-            sleep 1
-          done
-          kubectl exec -it $POD_NAME -- kubectl get crd appconfigs.stable.example.com websites.extensions.example.com
+          make test-example-apiserver
+      - name: Push images
+        timeout-minutes: 3
+        run: |
+          echo TODO add push credentials to repo
+          # make docker-push
+          # make -C test/apiserver push
```
kd7lxl, line 68: I would put this in a different workflow that only runs on `main`.

test/apiserver/makefile
kd7lxl: Since kind already checks to see if an image is already loaded, this optimization doesn't buy much.

.github/workflows/push.yaml
```diff
@@ -0,0 +1,29 @@
+name: PR
```
kd7lxl, line 1:
```suggestion
name: Push Images
```

.github/workflows/push.yaml
```diff
@@ -0,0 +1,29 @@
...
+on: pull_request
```
kd7lxl, line 2:
```suggestion
on: push
```

.github/workflows/push.yaml
```diff
@@ -0,0 +1,29 @@
...
+jobs:
+  push:
+    branchs:
+    - main
+    name: End To End
```
kd7lxl, line 7:
```suggestion
    name: Push Images
```

test/apiserver/makefile
drewwells: done

.github/workflows/pr.yaml
```diff
@@ -79,6 +67,17 @@ jobs:
...
+        timeout-minutes: 5
+        run: |
+          make deploy-example-apiserver KIND_NAME="chart-testing"
+          POD_NAME=$(kubectl get pod -o name -l app.kubernetes.io/name=konk-service )
+          until kubectl exec -it $POD_NAME -- kubectl get contact
+          do
+            kubectl exec -it $POD_NAME -- kubectl describe apiservice v1alpha1.example.infoblox.com
+            sleep 3
+          done
+          make test-example-apiserver
```
kd7lxl, line 80: Why do we run this target twice? The tests already take a long time.

.github/workflows/pr.yaml
```diff
@@ -36,41 +36,29 @@ jobs:
...
-          until cat test/apiserver/sample/contact.yaml | kubectl exec -i $POD_NAME -- kubectl apply -f -
-          do
-            sleep 1
-          done
-          until kubectl exec -it $POD_NAME -- kubectl get contact contact-example
-          do
-            sleep 1
-          done
-          kubectl exec -it $POD_NAME -- kubectl get crd appconfigs.stable.example.com websites.extensions.example.com
+          make test-example-apiserver
```
kd7lxl, line 61: The other instance.

.github/workflows/pr.yaml
drewwells: done

.github/workflows/pr.yaml
drewwells: ?

.github/workflows/pr.yaml
```diff
@@ -79,6 +67,17 @@ jobs:
           do
             sleep 3
           done
+      - name: Deploy/Test APIServer embedded KONK
```
abalaven, line 70: What's the difference between this and the previous test?

.github/workflows/push.yaml
```diff
@@ -0,0 +1,29 @@
+name: Push
+on: push
+jobs:
+  push:
+    branchs:
```
abalaven, line 5: Should be `branches`

.github/workflows/pr.yaml
drewwells: embedded KONK vs non-embedded KONK. I don't really have a way to cheat at this. Perhaps we deploy both examples at the same time 🤔

.github/workflows/push.yaml
drewwells: thanks, fixed

.github/workflows/pr.yaml
kd7lxl: In parallel independent workflows? SGTM.

.github/workflows/pr.yaml
drewwells: Removed

.github/workflows/pr.yaml
drewwells: All the example-apiserver tests should probably move to a separate workflow. I left out the embedded test for now.

helm-charts/example-apiserver/templates/konk.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: konk.infoblox.com/v1alpha1
+kind: Konk
+metadata:
+  name: {{ .Release.Name }}-konk
```
abalaven, line 4: I don't think we should include the "-konk" suffix whenever we create a konk. Its difficult to predict what the behavior would be if someone creates a konk without the "-konk" suffix.

helm-charts/konk-service/templates/deployment.yaml
```diff
@@ -62,7 +62,7 @@ spec:
       volumes:
       - name: kubeconfig
         secret:
-          secretName: {{ .Values.konk.name | required "must specify konk.name" }}-konk-kubeconfig
+          secretName: {{ .Values.konk.name | required "must specify konk.name" }}-kubeconfig
```
kd7lxl, line 65: I'm suspicious that this will work because I don't see a diff on the section that creates this secret.

helm-charts/example-apiserver/templates/konk.yaml
kd7lxl: We can't have duplicate release names on a cluster, so if they create a Konk and a KonkService with exactly the same name there will be a conflict. This is the benefit of including `-konk` in the name. I don't have a better solution.

If we want to enforce this, we could make it part of the CRD validation.

drewwells: the CR name becomes the helm release name. We need the suffix to ensure Konk release do not conflict with Konkservice releases

helm-charts/konk-service/templates/deployment.yaml
drewwells: this should be calling the helper

abalaven: This will work. Since `-konk` suffix is added while creating the konk, the kubeconfig secret is named as `runner-konk-kubeconfig`.

drewwells: Adding a test

helm-charts/konk-service/templates/_helpers.tpl
```diff
@@ -81,3 +81,16 @@ dummy
...
 {{- end -}}
+
+{{/*
+Templates konk name
+*/}}
+{{- define "konk-service.konkname" -}}
+{{- if hasSuffix "-konk" .Values.konk.name }}
+{{- .Values.konk.name }}
+{{- else }}
+{{- required "konk name must end in -konk" .Values.notexist }}
```
kd7lxl, line 92: This seems like the wrong place to enforce this. It won't fail until the operator tries to install the KonkService, and by then the Konk already exists. The better place to enforce this is the validation section of the CRD, since that will give immediate visible feedback to the user during installation.

Also, you can use `fail` for this:
```suggestion
{{- fail "konk name must end in -konk" }}
```

drewwells: good idea for the next PR. I wanted to enforce the idea that we're not going to attempt to fix the konk.name inside template code. I want it to fail.

.github/workflows/push.yaml
```diff
@@ -0,0 +1,30 @@
...
+jobs:
+  push:
+    name: Push Images
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    defaults:
+      run:
+        working-directory: ./src/github.com/${{ github.repository }}
+    env:
+      GOPATH: ${{ github.workspace }}
```
kd7lxl, line 15: Shouldn't be any need to `cd` into the `GOPATH`, since this job doesn't run `go`.

helm-charts/konk-service/templates/_helpers.tpl
kd7lxl: The problem is that the template failure gets buried in the CRD status message. It's not an obviously failure. Not very user friendly.

.github/workflows/push.yaml
drewwells: Actually the opposite, it needs the go action so the theoretical push command can build the images. Im going to remove the file so this PR can get approved some day

kd7lxl: The image we build now doesn't build any go... It just copies the base helm-operator image.

drewwells: example-apiserver has to build go code

kd7lxl (approved):
You've worked hard for this ⭐

abalaven (approved):

PR #88, opened 2020-11-04
Allow the user to provide their own APIService caBundle

kd7lxl:


abalaven:
Is there some way we can test this in the end2end workflow?

kd7lxl:
> Is there some way we can test this in the end to end workflow?

Not easily. It turns off a feature of KonkService that we want to test, and our test supports support running multiple cases. The test case with the feature is more important than the case that disables it, so I left the test as-is. It would be helpful to rewrite the test workflow in a way that supported multiple cases.

abalaven (approved):

PR #101, opened 2020-11-09
support creating cluster-scoped Konks

kd7lxl:
This adds a feature in Konk and KonkService to run the Konk in a different namespace than the KonkService. When Konk scope is set to `cluster`, it will install its CA in a `ClusterIssuer`. The KonkService will then use this `ClusterIssuer` to get a kubeconfig certificate signed and bundle that certificate in a kubeconfig in its own namespace.

~Todo~ done:
- [x] create CA secret in cert-manager namespace
- [x] create ClusterIssuer
- [x] use kubeconfig generated by cert-manager in konk-service


helm-charts/example-apiserver/templates/konkservice.yaml
```diff
@@ -26,5 +26,3 @@ spec:
   {{- with .Values.konkservice }}
   {{- toYaml . | nindent 2 }}
   {{- end }}
-  crds: |
-{{ (.Files.Glob "crds/*").AsConfig | indent 4 }}
```
drewwells, line 30: 👍

abalaven: We can also remove the `crds` directory as well.

kd7lxl: I think I'll move this to another PR to simplify this one.

https://github.com/infobloxopen/konk/pull/102

Makefile
```diff
@@ -217,7 +221,10 @@ deploy-ingress-nginx:
 		kubectl --namespace ingress-nginx describe pod -l app.kubernetes.io/component=controller; \
 	done
 
-deploy-apiserver: HELM_FLAGS ?=--set=image.tag=$(GIT_VERSION) --set=image.pullPolicy=IfNotPresent --set=konk.name=${KONK_NAME}
+ifndef KONK_NAME
+deploy-apiserver: CREATE_KONK ?= true
+endif
+deploy-apiserver: HELM_FLAGS ?=--set=image.tag=$(GIT_VERSION) --set=image.pullPolicy=IfNotPresent --set=konk.create=${CREATE_KONK} --set=konk.name=${KONK_NAME}
```
drewwells, line 227: 👍

kd7lxl: I could pull this out into its own PR too, if you want it sooner.

config/crd/bases/konk.infoblox.com_konks.yaml
```diff
@@ -35,11 +35,23 @@ spec:
             description: Spec defines the desired state of Konk
             type: object
             x-kubernetes-preserve-unknown-fields: true
+            properties:
+              scope:
+                description: Defines the scope of access to this konk
+                type: string
+                pattern: '^cluster|namespace$'
```
drewwells, line 42: is there a reason to ever namespace scope access? Are you thinking a namespace-secure konk? For instance, if you have rbac access to create a certificate, you can use any clusterissuer. It would simplify our test matrix to only think about cluster scoped

kd7lxl: Yeah, I figure there will be konk users who want it to be namespace-secure. I don't have a user for it now -- our users only want cluster-scoped, and I agree the test matrix is simpler if we only cluster-scope.

drewwells (approved):

watches.yaml
```diff
@@ -3,6 +3,8 @@
   version: v1alpha1
   kind: Konk
   chart: helm-charts/konk
+  overrideValues:
+    certManager.namespace: $CERT_MANAGER_NAMESPACE
```
drewwells, line 7: This will mean that users can not override the certManager.namespace

drewwells:
There's retries for the helm test calls now, but they passed on the first call in this PR https://github.com/infobloxopen/konk/runs/1381088761?check_suite_focus=true

watches.yaml
kd7lxl: Yes, the intent is that we set this for them in the operator, since it will be consistent across the cluster.

Makefile
kd7lxl: #111

drewwells:
image

.github/workflows/pr.yaml
```diff
@@ -96,7 +99,7 @@ jobs:
         run: |
           # deploy example-apiserver using a passed KONK reference
           make deploy-apiserver KIND_NAME="chart-testing" KONK_NAME="runner-konk"
-          POD_NAME=$(kubectl get pod -o name -l app.kubernetes.io/name=konk-service )
+          POD_NAME=$(kubectl get pod -o name -l app.kubernetes.io/name=konk-service,app.kubernetes.io/component=apiservice )
```
drewwells, line 102: If I understand this correctly, the apiserver gets installed into `default`. So moving konk into a different namespace tests cross-namespace support

.github/workflows/pr.yaml
```diff
@@ -16,6 +17,7 @@ jobs:
         working-directory: ./src/github.com/${{ github.repository }}
     env:
       GOPATH: ${{ github.workspace }}
+      KONK_NAMESPACE: konk
```
drewwells, line 20: call me paranoid, but I'd rather we test konk running in any namespace other than default or `konk`. This will separate any unintentional dependencies on resources in the operator namespace.

Moving the konk-operator to a different namespace would test the same thing.

helm-charts/example-apiserver/templates/_helpers.tpl
```diff
@@ -67,4 +67,14 @@ Templates konk name
...
+{{- end }}
+
+{{/*
+Templates konk-service name
+*/}}
+{{- define "example-apiserver.konk-service-name" -}}
+{{- .Release.Name | trunc 63 | trimSuffix "-" }}-api
+{{- end }}
+{{- define "example-apiserver.konk-service-fullname" -}}
+{{- include "example-apiserver.konk-service-name" . }}-konk-service
```
drewwells, line 79: `konk-service` 🤮

.github/workflows/pr.yaml
kd7lxl: Yes, you understood correctly. That's why the PR moves the konk into `konk` namespace.

helm-charts/example-apiserver/templates/_helpers.tpl
kd7lxl: contacting all our users for a secret name change 🤮

helm-charts/konk-operator/templates/deployment.yaml
```diff
@@ -35,6 +35,8 @@ spec:
           env:
             - name: AUTH_URL
               value: {{ .Values.authURL | default "" }}
+            - name: CERT_MANAGER_NAMESPACE
+              value: {{ (index .Values "cert-manager").namespace }}
```
drewwells, line 39: wouldn't it have been easier to remove the `-` ? I don't see any proprietary code defining "cert-manager" today.

helm-charts/konk/scripts/provision.sh
```diff
@@ -62,6 +62,17 @@ then
   kubectl -n $NAMESPACE label secret $FULLNAME-etcd-ca $LABELS
 fi
 
+if [ "$SCOPE" = "cluster" ]
+then
+  if secret_not_found $FULLNAME-ca $CERT_MANAGER_NAMESPACE
```
drewwells, line 67: Was there a technical requirement that the secret lived in cert-manager namespace? Can you document that or link to a jetstack issue describing the limitation?

.github/workflows/pr.yaml
kd7lxl: Moved to `baz`.

helm-charts/konk-operator/templates/deployment.yaml
kd7lxl: When installed as a subchart, cert-manager's values are under `cert-manager`.

helm-charts/konk/scripts/provision.sh
kd7lxl: https://docs.cert-manager.io/en/release-0.11/reference/clusterissuers.html
> <p>When referencing a <code class="docutils literal notranslate"><span class="pre">Secret</span></code> resource in <code class="docutils literal notranslate"><span class="pre">ClusterIssuer</span></code> resources (eg <code class="docutils literal notranslate"><span class="pre">apiKeySecretRef</span></code>) the <code class="docutils literal notranslate"><span class="pre">Secret</span></code> needs to be in the same namespace as the <code class="docutils literal notranslate"><span class="pre">cert-manager</span></code> controller pod. You can optionally override this by using the <code class="docutils literal notranslate"><span class="pre">--cluster-resource-namespace</span></code> argument to the controller.</p>

drewwells (approved):

PR #109, opened 2020-11-10
remove exit on error

### LINKED ISSUE #103: Flakey test
The test has been failing here occasionally, even though the output looks successful:
https://github.com/infobloxopen/konk/pull/101/checks?check_run_id=1380932845

drewwells:
fixes #103


kd7lxl (approved):
So the only way to fail now is timeout? Okay.

drewwells:
> So the only way to fail now is timeout? Okay.

We may need a better readiness check so we don't rely on infinite retries for this test to pass... but I still don't see it failing on the first call on main.

Issue #113, opened 2020-11-11
Konk CA should be managed by CertManager


drewwells:
Provision script is creating brand new CA/Etcd CA on every start. It then walks through the k8s secrets to see if any are missing and replaces them. If it happens to replace a secret, it is most likely to break client-server communication due to CA being generated on every start.

We should manage CA as separate CertManager owned certificates. They can share an issuer, but Etcd and K8s should use different CAs. The init pod will mount these CAs and be read by kubeadm to create client/server certificates.

Kubeadm may not have flags, but has predefined file locations where the CA should be found. Mount the secret in these locations.
```
kubeadm init phase certs etcd-server                                                            stg-1
W1111 16:24:06.679820 2604580 configset.go:348] WARNING: kubeadm cannot validate component configs for API groups [kubelet.config.k8s.io kubeproxy.config.k8s.io]
error execution phase certs/etcd-server: couldn't load CA certificate etcd-ca: couldn't load etcd/ca certificate authority from /etc/kubernetes/pki
To see the stack trace of this error execute with --v=5 or higher
```

- [x] k8s ca survives pod restarts k8s secret storage
- [ ] etcd ca survives pod restarts k8s secret storage

kd7lxl:
I partially fixed this issue here: https://github.com/infobloxopen/konk/pull/101#pullrequestreview-528592356

PR #114, opened 2020-11-12
Load existing CA

kd7lxl:
Broken out from #101, this fix loads an existing CA secret before running `kubeadm`, so that certs generated by `kubeadm` are signed by the same original CA and trusted by all the user pods.

### DIFF
--- a/helm-charts/konk/scripts/provision.sh
+++ b/helm-charts/konk/scripts/provision.sh
@@ -1,6 +1,16 @@
 #!/bin/bash
 set -xe
 
+# load existing CA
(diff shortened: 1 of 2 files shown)


helm-charts/konk/scripts/provision.sh
```diff
@@ -1,6 +1,16 @@
 #!/bin/bash
 set -xe
 
+# load existing CA
```
kd7lxl, line 4: As part of #113, `/etc/kubernetes/pki/etcd/ca.crt` should also be loaded from an existing secret when possible.

abalaven (approved):

PR #115, opened 2020-11-12
konk and example-apiserver smoke test

drewwells:


scripts/konk-smoke-test.sh
```diff
@@ -0,0 +1,22 @@
...
+trap cleanup EXIT
+
+kubectl create ns konk-test || true
+kubectl -n konk-test apply -f ../examples/konk.yaml
+
+kubectl wait --timeout=1s --for=condition=Deployed -n konk-test konk runner-konk
+
+REASON=$(kubectl -n konk-test get konk runner-konk -o jsonpath="{range.status.conditions[?(@.type=='Deployed')]}{$@.type}={@.reason}{'\n'}{end}")
+
+if [[ $REASON != *"InstallSuccessful" ]]; then
```
kd7lxl, line 18: add a case for `UpgradeSuccessful`.

```
% k get konks -A                                                                                                                                    env-4
NAMESPACE    NAME                         STATUS              AGE
aggregate    tagging-aggregate-api-konk   InstallSuccessful   46h
pferrell     konk2-konk                   UpgradeSuccessful   7d23h
tagging-v2   tagging-aggregate-api        InstallSuccessful   7d23h
thayward     example-konk                 UpgradeSuccessful   36d
```

kd7lxl: Oh, I see you want to always start clean, not test upgrade. In this case checking for `UpgradeSuccessful` is not required, but also missed the common smoke case...

Maybe another test should check that no existing konks have failed to upgrade.

drewwells: I'm hesitant that an upgrade fully tests creation functionality. The test also does a full teardown regardless if it passes or succeeds

kd7lxl (approved):

scripts/konk-smoke-test.sh
```diff
@@ -0,0 +1,22 @@
...
+trap cleanup EXIT
+
+kubectl create ns konk-test || true
+kubectl -n konk-test apply -f ../examples/konk.yaml
+
+kubectl wait --timeout=1s --for=condition=Deployed -n konk-test konk runner-konk
+
+REASON=$(kubectl -n konk-test get konk runner-konk -o jsonpath="{range.status.conditions[?(@.type=='Deployed')]}{$@.type}={@.reason}{'\n'}{end}")
+
+if [[ $REASON != *"Successful" ]]; then
```
kd7lxl, line 18: As you mentioned before, this tests the wrong thing because you expect `runner-konk` to be a clean install, not an upgrade. If it's being upgraded, something went wrong with the cleanup stage.

drewwells, line 18:
```suggestion
if [[ $REASON != *"InstallSuccessful" ]]; then
```

PR #116, opened 2020-11-12
enable vendoring with dependabot

kd7lxl:
Just testing to see if this passes the schema check.


kd7lxl:
Hmm, there was no Dependabot build check. Maybe that's a beta-only feature.

PR #120, opened 2020-11-13
Add etcd ca cert

### LINKED ISSUE #113: Konk CA should be managed by CertManager

abalaven:
Fixes #113 

```
NAME                                               TYPE                                  DATA   AGE
runner-konk-etcd-ca                                kubernetes.io/tls                     2      2m29s
```

### DIFF
--- a/helm-charts/konk/scripts/provision.sh
+++ b/helm-charts/konk/scripts/provision.sh
@@ -1,6 +1,15 @@
 #!/bin/bash
 set -xe
 
+secret_not_found () {
+  if [ "$(kubectl -n $NAMESPACE get secret $1 --ignore-not-found 2>&1 )" = "" ]
+  then
+    return 0
+  else
+    return 1
+  fi
+}
+
 # load existing CA
 if kubectl -n $NAMESPACE get secret $FULLNAME-ca
 then
@@ -22,7 +31,7 @@ fi
 kubeadm init phase certs all --apiserver-cert-extra-sans $FULLNAME,$FULLNAME.$NAMESPACE,$FULLNAME.$NAMESPACE.svc,$FULLNAME.$NAMESPACE.svc.cluster.local
 kubeadm init phase kubeconfig admin --control-plane-endpoint $FULLNAME.$NAMESPACE.svc
 find /etc/kubernetes/pki
-if ! kubectl -n $NAMESPACE get secret $FULLNAME-etcd-cert
+if secret_not_found $FULLNAME-etcd-cert
 then
   kubectl -n $NAMESPACE create secret generic $FULLNAME-etcd-cert \
     --from-file=/etc/kubernetes/pki/etcd/ca.crt \
@@ -30,7 +39,7 @@ then
     --from-file=/etc/kubernetes/pki/etcd/server.key
   kubectl -n $NAMESPACE label secret $FULLNAME-etcd-cert $LABELS
 fi
-if ! kubectl -n $NAMESPACE get secret $FULLNAME-apiserver-cert
+if secret_not_found $FULLNAME-apiserver-cert
 then
   kubectl -n $NAMESPACE create secret generic $FULLNAME-apiserver-cert \
     --from-file=/etc/kubernetes/pki/apiserver.crt \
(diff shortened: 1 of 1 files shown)


helm-charts/konk/scripts/provision.sh
```diff
@@ -40,6 +48,13 @@ then
     --key=/etc/kubernetes/pki/ca.key
   kubectl -n $NAMESPACE label secret $FULLNAME-ca $LABELS
 fi
+if ! kubectl -n $NAMESPACE get secret $FULLNAME-etcd-ca
```
drewwells, line 51: we need to check for 'not found' errors specifically. We could run into a scenario where a 503 service unavailable causes the secret to get overridden

abalaven: Makes sense. That would also be the case for the k8s CA as well, right?

kd7lxl: This logic might work for you. Experiment with it...
```suggestion
if [ "$(kubectl -n $NAMESPACE get secret $FULLNAME-etcd-ca --ignore-not-found 2>&1 )" = "" ]
```
It checks that the response is successful, but empty, which is what you would expect from this command when a secret does not exist.

helm-charts/konk/scripts/provision.sh
```diff
@@ -41,14 +41,14 @@ then
     --from-file=/etc/kubernetes/pki/apiserver-etcd-client.key
   kubectl -n $NAMESPACE label secret $FULLNAME-apiserver-cert $LABELS
 fi
-if ! kubectl -n $NAMESPACE get secret $FULLNAME-ca
+if [ "$(kubectl -n $NAMESPACE get secret $FULLNAME-ca --ignore-not-found 2>&1 )" = "" ]
```
kd7lxl, line 44: We actually need this technique for every place `get secret` is called. The cleanest way to do it would be to write a function.

helm-charts/konk/scripts/provision.sh
```diff
@@ -1,6 +1,15 @@
...
+  if [ "$(kubectl -n $NAMESPACE get secret $1 --ignore-not-found 2>&1 )" = "" ]
+  then
+    return 0
+  else
+    return 1
+  fi
+}
+
 # load existing CA
 if kubectl -n $NAMESPACE get secret $FULLNAME-ca
```
kd7lxl, line 14: Use it here too. If `kubectl` returns an error, this should retry or abort rather than skipping loading the CA cert.

helm-charts/konk/scripts/provision.sh
```diff
@@ -1,6 +1,15 @@
 #!/bin/bash
 set -xe
 
+secret_not_found () {
+  if [ "$(kubectl -n $NAMESPACE get secret $1 --ignore-not-found 2>&1 )" = "" ]
+  then
+    return 0
+  else
+    return 1
+  fi
```
kd7lxl, line 10: We can make this a little better. Right now, it's short and sweet but completely hides errors. In the function, we have the opportunity to be more verbose even though it wasn't feasible as a one-liner.

```suggestion
  output=$(kubectl -n $NAMESPACE get secret $1 --ignore-not-found 2>&1 )
  echo "$output"
  [ "$output" = "" ]
```
BTW, `return` is implicit in my example:
https://www.gnu.org/software/bash/manual/bash.html#Shell-Functions
> otherwise the function’s return status is the exit status of the last command executed

helm-charts/konk/scripts/provision.sh
```diff
@@ -1,28 +1,42 @@
 #!/bin/bash
 set -xe
 
+secret_not_found () {
+  output=$(kubectl -n $NAMESPACE get secret $1 --ignore-not-found 2>&1 )
+  echo "$output"
```
kd7lxl, line 6: The other possibility I see for improvement here would be to retry with backoff and eventually abort (`exit 1`) upon error. This way it would be eventually consistent during a kube-apiserver outage.

kd7lxl (approved):
Original scope of the PR has been achieved.

PR #132, opened 2020-11-20
Update readme with secret usage

kd7lxl:
This adds a Usage section to the readme. Here I've included what's basically a FAQ. The two sections answer the most basic "how do I use konk?" questions. They document the naming convention for the secrets, which they will need to know to use certificates generated by konk/konkservice in their pods.


drewwells (approved):
I miss the helm chart values style documentation, but this is great too. Thanks

PR #135, opened 2020-11-25
 use etcd operator in konk chart for high availability

kd7lxl:
For high availability, we need multiple etcd pods with durable storage. The bitnami/etcd chart provides that. #2 experimented with running that chart as a subchart, but there was a major caveat in that it required obtuse configuration by the user. This PR explores deploying the same chart with helm-operator. Now, konk deploys and uses an `Etcd` resource by default, created by konk-operator from the bitnami/etcd chart.

Unfortunately the operator-sdk helm-operator only works with untarred charts, so the bitnami/etcd chart had to be cloned rather than included as a tarball. I also included a make target to clone/upgrade it. `helm-charts/etcd/` should be treated as generated files and not modified except with this make target.

Replaces #2


drewwells:
You could submodule their repo and write something that copies the charts we want into helm-charts/etcd. Submodules are not easy to work with. 

Is it possible to write an empty chart that subcharts etcd? [:ticket:](https://github.com/operator-framework/operator-sdk/issues/new?assignees=&labels=&template=bug-report.md&title=)

kd7lxl:
> You could submodule their repo and write something that copies the charts we want into helm-charts/etcd. Submodules are not easy to work with.
> 
> Is it possible to write an empty chart that subcharts etcd? [🎫](https://github.com/operator-framework/operator-sdk/issues/new?assignees=&labels=&template=bug-report.md&title=)

I'm not sure this suggestion actually adds value. The way it is, the chart contents are easily readable and easy to update/sync with the upstream chart. The only caveat is that it's not obvious the code is generated / may be overwritten.

Makefile
```diff
@@ -45,6 +45,11 @@ deploy-cert-manager:
 		--set installCRDs=true \
 		--set extraArgs[0]="--enable-certificate-owner-ref=true""
 
+deploy-crds: konk-operator-${GIT_VERSION}.tgz
+	# https://helm.sh/docs/chart_best_practices/custom_resource_definitions/
+	# > There is no support at this time for upgrading or deleting CRDs using Helm.
+	kubectl apply -f helm-charts/konk-operator/crds
+
```
drewwells, line 52: 👍

helm-charts/konk/templates/deployment.yaml
```diff
@@ -124,6 +130,7 @@ spec:
           - mountPath: /etc/kubernetes/pki/apiserver/etcd
             name: etcd-client-certs
             readOnly: true
+        {{- end }}
```
drewwells, line 133: do tests exercise this version of installing etcd as a sidecar in the pod?

helm-charts/konk/templates/etcd.yaml
```diff
@@ -0,0 +1,27 @@
...
+spec:
+  auth:
+    rbac:
+      enabled: false
+    client:
+      secureTransport: true
+      existingSecret: {{ include "konk.fullname" $ }}-etcd-cert
+      certFilename: server.crt
+      certKeyFilename: server.key
+  # these probes don't work with self-signed certs because probe.sh doesn't set the cacert path in the $AUTH_OPTIONS
```
drewwells, line 19: Probably worth mentioning in the comments that readiness probes don't work well with Etcd. For cluster initialization, the pods need to accept traffic prior to being ready to do initial gossip. They can talk to each other on pod IPs, but that's a bit fragile.
https://github.com/improbable-eng/etcd-cluster-operator/blob/master/docs/operations.md#frequently-asked-questions

drewwells (approved):
Thanks, this looks good

helm-charts/konk/templates/deployment.yaml
kd7lxl: No, rather than increasing the test matrix I just rewrote the test to use the etcd operator. We could fairly easily add this test back by supplying two values files as input to the test matrix. Naturally, that doubles the number of tests we run.

PR #145, opened 2020-12-16
Verify proxy certs

drewwells:
- [x] make deploy-example-apiserver just works
- [x] didn't break shared konk (build checks this)
- [ ] verify proxy-client-certificates are valid as part of test
- [x] set admin.conf kubeconfig's owner to kubeconfig deployment


drewwells:
DEMO
client certificates do work now
```
openssl verify -CAfile <(k -n default get secret -o yaml drew-apiserver-konk-proxy-client | awk '/^  ca.crt/ {print $2}' | base64 -d) <(k -n default get secret -o yaml drew-apiserver-konk-proxy-client | awk '/^  tls.crt/ {print $2}' | base64 -d)                             kind-minikube
/proc/self/fd/18: OK
```

helm-charts/konk-service/templates/kubeconfig-rbac.yaml
```diff
@@ -17,6 +17,12 @@ rules:
     - list
     - patch
     - watch
+  - apiGroups:
+    - "apps"
+    resources:
+    - deployments
+    verbs:
+    - get
```
drewwells, line 25: kubeconfig generating pod now inspects its deployment to set owner UID.

helm-charts/example-apiserver/templates/_helpers.tpl
```diff
@@ -63,18 +63,30 @@ Create the name of the service account to use
...
+{{- .Values.konkservice.name | default (printf "%s-konk-service" .Release.Name) }}
 {{- end }}
-{{- define "example-apiserver.konk-service-fullname" -}}
-{{- include "example-apiserver.konk-service-name" . }}-konk-service
+
+{{/*
+Templates konk-service namespace
+*/}}
+{{- define "example-apiserver.konk-service-namespace" -}}
+{{- .Values.konkservice.namespace | default .Release.Namespace }}
```
kd7lxl, line 91: I can't think of any case where this could differ from `.Release.Namespace`. If it does, the secrets aren't readable by the app.

helm-charts/konk-service/values.yaml
```diff
@@ -51,6 +51,7 @@ ingress:
   #      - chart-example.local
 
 konk:
+  create: true
```
kd7lxl, line 54: Changing this default will create a bunch of konks on existing services that are currently sharing one.

.github/workflows/pr.yaml
```diff
@@ -90,24 +89,38 @@ jobs:
           done
           make test-konk KONK_NAMESPACE=$KONK_NAMESPACE
           make test-konk-local
+
+          until kubectl wait -n $KONK_NAMESPACE --timeout=1m --for=condition=ready certificate -l component=proxy-client,app.kubernetes.io/name=konk
+          do
+            sleep 1
+          done
```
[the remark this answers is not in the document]
drewwells: Removing, I don't think this is needed. The timing issues were coming from building go binaries in container

.github/workflows/pr.yaml
```diff
@@ -90,24 +89,38 @@ jobs:
...
           go-version: 1.15.x
-      - name: Test Extension APIServer
+      - name: Build Extension APIServer
         timeout-minutes: 7
         run: |
+          make kind-load-apiserver KIND_NAME="chart-testing"
+      - name: Deploy Extension APIServer
+        timeout-minutes: 4
+        run: |
+          kubectl
```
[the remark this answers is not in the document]
drewwells: removed

kd7lxl (approved):

helm-charts/konk-service/values.yaml
```diff
@@ -51,6 +51,7 @@ ingress:
   #      - chart-example.local
 
 konk:
+  create: false
```
kd7lxl, line 54: I don't think this is actually read anywhere.

helm-charts/konk-service/values.yaml
```diff
@@ -53,6 +53,7 @@ ingress:
 konk:
   # The value below should be overriden by the konk-name
   name: ""
+  namespace: ""
```
drewwells, line 56: DEMO
Taken from chart, unless specified
```
helm template --set ingress.enabled=true --api-versions cert-manager.io/v1alpha3 helm-charts/konk-service | grep proxy-ssl-secret
    nginx.ingress.kubernetes.io/proxy-ssl-secret: default/-ingress-client
```

```
helm template --set konk.namespace=baz --set ingress.enabled=true --api-versions cert-manager.io/v1alpha3 helm-charts/konk-service | grep proxy-ssl-secret
    nginx.ingress.kubernetes.io/proxy-ssl-secret: baz/-ingress-client
```

kd7lxl: This is fine. It was implicitly `null` before.

kd7lxl (approved):

PR #149, opened 2020-12-17
add space

kd7lxl:
Removes the requirement for users to define their own `imagePullSecrets` in the CRs when space controller is present.


drewwells (changes requested):
konkoperator, konkservice, example-apiserver SA needs to reference this optional secret

drewwells (approved):

PR #156, opened 2021-01-13
Test delete of konk and konkservice resources

kd7lxl:


.github/workflows/pr.yaml
```diff
@@ -123,6 +123,10 @@ jobs:
           do
             sleep 3
           done
+      - name: Test delete
+        timeout-minutes: 1
+        run: |
+          kubectl delete konks,konkservices --all --all-namespaces
```
drewwells, line 129: We should check with helm that the operator actually completed the task.
```suggestion
          kubectl delete konks,konkservices --all --all-namespaces
          helm -n $KONK_NAMESPACE | grep -c runner-konk 
```

drewwells: might need to add `set -e` to get github-actions to fail on matches

kd7lxl: `set -e` is default in actions.

kd7lxl: Added support for this.

drewwells (approved):

PR #161, opened 2021-01-21
drop Tiller support

kd7lxl:
`helm template --set=crds.create=true` with helm3 doesn't work with CRDs that have these `crd-install` hooks. Since `crd-install` was EOL with helm2, we can remove it.


drewwells (approved):
good bye helm2

PR #187, opened 2021-02-27
Add a smoke test to konkservice

kd7lxl:
This new deployment is part of konkservice and periodically attempts to list resource types from the apiserver. If the apiserver doesn't respond properly, the pod indicates unready. This could be used in conjunction with tools like `kubectl wait --for=condition=ready pod -l app.kubernetes.io/component=apiservice-test` for troubleshooting or kube-state-metrics for monitoring and alerting.

### DIFF
--- a/helm-charts/konk-service/templates/apiservice-test-deployment.yaml
+++ b/helm-charts/konk-service/templates/apiservice-test-deployment.yaml
@@ -1,0 +1,63 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "konk-service.fullname" . }}-kubectl-apiservice-test
+  namespace: {{ .Release.Namespace }}
+  labels:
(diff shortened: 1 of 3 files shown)


helm-charts/konk-service/templates/apiservice-test-deployment.yaml
```diff
@@ -0,0 +1,63 @@
...
+          {{- toYaml .Values.kind.resources | nindent 10 }}
+        volumeMounts:
+        - name: kubeconfig
+          mountPath: "/etc/kubernetes"
+          readOnly: true
+        - name: tmp
+          mountPath: "/tmp"
+        readinessProbe:
+          exec:
+            command: ["/bin/bash", "-c", "exit $(</tmp/healthy)"]
```
drewwells, line 55: Clever, very nice

drewwells (approved):

PR #193, opened 2021-03-18
replace kubeconfig after cert changes. Fixes #192

### LINKED ISSUE #192: Generate new kubeconfig before expiration
The certs used in the kubeconfig in konk-services are maintained/renewed by cert-manager and default to a 90-day expiration. However, after using them to build the kubeconfig, konk-service never replaces them with the renewed certificate. After 90 days, the apiservice fails with certificate errors. After a cert is renewed, konk-service should rebuild and replace the kubeconfig secret.
https://github.com/infobloxopen/konk/blob/297208aa954a8a54c4b7ac1aa700a1be3878029a/helm-charts/konk-service/templates/kubeconfig-deployment.yaml#L52

https://infoblox.atlassian.net/browse/ATLAS-9340

kd7lxl:
This change will cause the konk-service-kubeconfig secret to be rebuilt and replaced in the following conditions:
* on konk-service-kubeconfig pod startup (e.g., new konk service created or konk-operator upgraded)
* within 3 minutes of cert-manager renewing the kubeconfig cert

### DIFF
--- a/helm-charts/konk-service/templates/kubeconfig-deployment.yaml
+++ b/helm-charts/konk-service/templates/kubeconfig-deployment.yaml
@@ -51,6 +51,8 @@ spec:
               unset KUBECONFIG
               if [ "$(kubectl -n $NAMESPACE get secret {{ include "konk-service.fullname" . }}-kubeconfig --ignore-not-found 2>&1 )" = "" ]
               then
+              (
+                set -e
                 kubectl -n $NAMESPACE create secret generic {{ include "konk-service.fullname" . }}-kubeconfig \
                   --from-file=$KUBECONFIG_PATH
                 kubectl -n $NAMESPACE label secret {{ include "konk-service.fullname" . }}-kubeconfig $LABELS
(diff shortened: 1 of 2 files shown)


helm-charts/konk-service/templates/kubeconfig-deployment.yaml
```diff
@@ -47,19 +47,28 @@ spec:
...
               kubectl config use-context {{ .Values.konk.name }}
 
-              # these commands run in infra cluster
+              # the following commands run in infra cluster
               unset KUBECONFIG
+
+              # deletes existing kubeconfig after a cert change
+              if ! md5sum --status --check /tmp/certsum
+              then
+                kubectl -n $NAMESPACE delete secret {{ include "konk-service.fullname" . }}-kubeconfig
```
kd7lxl, line 56: Will replacing the secret with a different one with the same name work for pods that already have it mounted? Or should I patch the existing one instead?

kd7lxl: Fixed.

abalaven (approved):

PR #213, opened 2021-05-17
pass kube-score

drewwells:
DEMO
The failing pieces are defined in DC or not enforced in our configuration.
image

### DIFF
--- a/helm-charts/konk-operator/values.yaml
+++ b/helm-charts/konk-operator/values.yaml
@@ -28,16 +28,18 @@ space:
 
 podAnnotations: {}
 
-podSecurityContext: {}
-  # fsGroup: 2000
+podSecurityContext:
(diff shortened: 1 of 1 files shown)


kd7lxl (approved):
I'll let the tests decide whether this works -- it's possible the operator-sdk image only works with particular user ids.

kube-score only does static analysis on the static parts of the manifest. This PR would be better if it also implemented the same practices on the konks and konkservices deployed by the operator.

drewwells:
> I'll let the tests decide whether this works -- it's possible the operator-sdk image only works with particular user ids.
> 
> kube-score only does static analysis on the static parts of the manifest. This PR would be better if it also implemented the same practices on the konks and konkservices deployed by the operator.

I'm just shooting for ✅  on DC builds right now

PR #216, opened 2021-05-25
Support Konk generates cert-manager in Minikube

hda2:
Northstar team wants to integrate Konk into minikube environment. We need to ensure the `cert-manager` is talking to minikube cluster and not other environments in the kube config file. 

This PR is to support Konk generates the cert-manager in Minikube by using `make deploy-cert-manager-minikube` command. 

The change was verified in the Northstar Minikube env. The cert manager was installed successfully.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 CHART_DIR	:= helm-charts
 GIT_VERSION	?= $(shell git describe --dirty=-unsupported --always --long --tags)
 HELM_IMAGE	?= infoblox/helm:3.2.4-5b243a2
+MINIKUBE        := $(HOME)/.minikube
 DOCKER_RUNNER	?= docker run --rm -i \
 			--entrypoint="" \
 			--network host \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -45,6 +59,14 @@ deploy-cert-manager:
 		--set installCRDs=true \
 		--set extraArgs[0]="--enable-certificate-owner-ref=true""
 
+# Run this only if your minikube does not have cert-manager already deployed
+deploy-cert-manager-minikube:
+	$(HELM_CMD_MINI) "helm repo add jetstack https://charts.jetstack.io && helm upgrade -i --wait cert-manager --namespace cert-manager jetstack/cert-manager --version v1.0.1 \
+		--create-namespace \
+		--kube-context minikube \
+		--set installCRDs=true \
+		--set extraArgs[0]="--enable-certificate-owner-ref=true""
```
kd7lxl, line 68: This duplicates existing functionality. I don't understand why the existing commands aren't working for your use case.

Officially, we support testing in `kind`. That works like this:
```
make kind
make deploy-cert-manager deploy-ingress-nginx
make kind-load-konk deploy-konk-operator
```

The equivalent should work for minikube:
```
minkube start
make deploy-cert-manager deploy-ingress-nginx
make kind-load-konk deploy-konk-operator
```
What happens when you run these commands?

hda2: after `minikube start`, when running `make deploy-cert-manager`, the following errors came out

```
Error: Kubernetes cluster unreachable: invalid configuration: [unable to read client-cert /Users/hdu/.minikube/client.crt for minikube due to open /Users/hdu/.minikube/client.crt: no such file or directory, unable to read client-key /Users/hdu/.minikube/client.key for minikube due to open /Users/hdu/.minikube/client.key: no such file or directory, unable to read certificate-authority /Users/hdu/.minikube/ca.crt for minikube due to open /Users/hdu/.minikube/ca.crt: no such file or directory]
make: *** [deploy-cert-manager] Error 1
```

drewwells (changes requested):
Please open a ticket with issue you are seeing with minikube. While we'd like to support every possible special k8s environment, it's not feasible to maintain such code.  You also don't need special commands to work with minikube. Minikube, KIND, k8s do not require special steps with konk as we do all of our installation via kubectl/helm.

You can follow how we communicate with kind in the integration tests. Those steps should work just as well with minikube. https://github.com/infobloxopen/konk/blob/main/.github/workflows/pr.yaml#L59-L62 

If you have not tried KIND, I'd highly recommend it. All of atlas has moved off of minikube to KIND. It is faster and does not require a VM.

Makefile
drewwells: embed certificates into the kubeconfig, some faq docs on this here: https://github.com/getporter/porter/pull/664/files#diff-83cea8d3073291cd4afd5c393ef0060b7fdc246e13eebf62029399d3556b9e92

drewwells: We use a docker container to call helm which is what is causing you issues. You could turn it off `DOCKER_RUNNER=helm make deploy-konk-operator`

hda2:
> If you have not tried KIND, I'd highly recommend it. All of atlas has moved off of minikube to KIND. It is faster and does not require a VM.

Thanks @drewwells and @kd7lxl , I tried `make kind`, then `make deploy-cert-manager` works. I will close this PR.

PR #218, opened 2021-05-28
preserve all headers with X- (to support passthrough of tracing headers)

pjferrell:


hramazani (approved):

kd7lxl (commented):
This is changing one of the defaults in the Konk spec. Do you have a demo of it working (by setting the field in the CR, not as a default)? Do you think this option is going to be popular enough to warrant making it the default?

pjferrell:
@kd7lxl I expect several projects leveraging konk will need to preserve X-Request-ID and X-B3-* headers for logging consistency and tracability.  Having it enabled as default will make it easier for users not familiar with the apiserver command-line flags.
I will deploy konk tomorrow, but don't expect any issue based on prior experience.  Will reach out to you on slack to discuss.

drewwells (commented):
This area needs tests to verify functionality. The defaults are the recommended defaults from upstream that you can override. Can you write a test verifying this actually works?

pjferrell:
Closing since @kd7lxl stated the Konk CR permitted the arguments to be specified in spec.  Documentation doesn't provide an example or any description of available fields in Konk CRD.

kd7lxl:
> Documentation doesn't provide an example or any description of available fields in Konk CRD.

Addressed in https://github.com/infobloxopen/konk/pull/223

PR #223, opened 2021-06-09
readme: reference Konk spec

kd7lxl:


hramazani:
See this PR as an example:
https://github.com/Infoblox-CTO/atlas.bulk/pull/135/files

PR #224, opened 2021-06-09
support arbitrary annotations on the APIService created by KonkService

kd7lxl:
https://infoblox.atlassian.net/browse/ATLAS-10229 requests a feature to add annotations to the `APIService` created by `KonkService`.


drewwells (approved):
Probably a good idea to allow editing labels too. Surprised we have no default labels, helm operator may add some by default.

PR #225, opened 2021-06-14
Update example

hramazani:


examples/konk.yaml
```diff
@@ -3,4 +3,8 @@ kind: Konk
 metadata:
   name: runner-konk
 spec:
+  spec:
+    apiserver:
+      remoteHeaders:
+        requestheader-extra-headers-prefix: "X-"
```
kd7lxl, line 9: I thought testing identified that this breaks authentication? Is it really okay to add to the standard example?

hramazani: Let's see how the authentication issue is being resolved, then I'll either close or update this PR.

PR #230, opened 2021-07-01
fix ci by using stable ingress-nginx

kd7lxl:
Related ticket: https://github.com/kubernetes-sigs/kind/pull/2347

Note: upgrade test cases fail because they use the ingress-nginx version from `main`, which is unstable.


hramazani (approved):

drewwells (approved):
k8s gathers version strings like this. It's kind of gross but what can you do.

PR #232, opened 2021-07-09
dashboard: alert on not-ready konks

kd7lxl:


helm-charts/konk-operator/dashboards/konk-operator.json
```diff
@@ -251,12 +265,57 @@
...
     {
+      "alert": {
+        "alertRuleTags": {},
+        "conditions": [
+          {
+            "evaluator": {
+              "params": [
+                0
+              ],
+              "type": "lt"
```
Calebjh, line 276: Is this checking for < 0 or <=0? Or is that not how the rule is read?

kd7lxl: `lt` means less than 0. The query returns -1 for not-ready pods, which will trigger this alert.

Calebjh (approved):

PR #240, opened 2021-08-16
memlimit for kubectl pods: limit memory to allow linux to free file cache

pjferrell:
Allocated memory grows due to linux file cache if no memory pressure/limit is configured.  This commit sets a default limit of 128Mi to allow linux to periodically free file cache (dentries/inodes) when utilization approaches limit.

### DIFF
--- a/helm-charts/konk-service/values.yaml
+++ b/helm-charts/konk-service/values.yaml
@@ -26,17 +26,15 @@ kind:
     pullPolicy: Always
     # -- Overrides the image tag whose default is the chart appVersion.
     tag: v1.19.0
-  resources: {}
+  resources:
     # We usually recommend not to specify default resources and to leave this as a conscious
     # choice for the user. This also increases chances charts run on environments with little
     # resources, such as Minikube. If you do want to specify resources, uncomment the following
     # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
-    # limits:
-    #   cpu: 100m
-    #   memory: 128Mi
-    # requests:
-    #   cpu: 100m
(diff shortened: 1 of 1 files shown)


pjferrell:
Changes to default resources was rejected by reviewers.  Each user deploying konk will need to override the resources to reduce memory held by kubectl containers.

pjferrell:
Reopened based on slack discussion with @drewwells  / @kd7lxl.

helm-charts/konk-service/values.yaml
```diff
@@ -26,17 +26,15 @@ kind:
...
-    # limits:
-    #   cpu: 100m
-    #   memory: 128Mi
-    # requests:
-    #   cpu: 100m
-    #   memory: 128Mi
+    limits:
+      memory: 128Mi
+    requests:
+      memory: 25Mi
```
kd7lxl, line 37: If we need to override this, like to test on a test-runner that has low resources, the ideal way would be to set `resources: {}` to remove them for the test. Unfortunately, a helm bug prevents this: https://github.com/helm/helm/issues/9136 I don't have a solution for this that allows setting a default other than `{}`.

pjferrell: @kd7lxl Is there a change that you're proposing or is this blocking your approval of this PR?

kd7lxl: It's blocking unless you know (or can research) how we can work around this helm bug. I want to maintain the capability of disabling the resource definitions, like when testing in kind. With the current proposal and bug I don't see a way to do that.

PR #248, opened 2021-09-10
upgrade cert-manager for k8s v1.22 compatibility

kd7lxl:
cert-manager support matrix: https://cert-manager.io/docs/installation/supported-releases/#supported-releases

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ helm-lint-%:
 
 # Run this only if your cluster does not have cert-manager already deployed
 deploy-cert-manager:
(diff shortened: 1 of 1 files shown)


kd7lxl:
```
W0910 20:40:55.580911       1 warnings.go:70] networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
```
Looks like #233 is also required to fix v1.22 support.

hramazani (approved):

PR #302, opened 2022-05-02
renew apiserver and etcd certs on konk pod startup

kd7lxl:
The etcd and apiserver certs generated by kubeadm are only valid for one year. Every time kubeadm runs, it will renew the certs, but before this change konk would never overwrite the existing certs with the new ones. This PR patches the existing certs with the freshly generated ones when the pod starts and every 90 days.

An alternative approach would be to redesign this process to use cert-manager (#303).

### DIFF
--- a/helm-charts/konk/scripts/provision.sh
+++ b/helm-charts/konk/scripts/provision.sh
@@ -41,30 +41,38 @@ kubeadm init phase certs etcd-server --config=/tmp/kubeadmcfg.yaml
 kubeadm init phase kubeconfig admin --control-plane-endpoint $FULLNAME.$NAMESPACE.svc
 find /etc/kubernetes/pki
 
-# replaces any existing etcd-cert
-if ! secret_not_found $FULLNAME-etcd-cert
+if secret_not_found $FULLNAME-etcd-cert
 then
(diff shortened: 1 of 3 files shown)


helm-charts/konk/scripts/provision.sh
```diff
@@ -40,25 +40,32 @@ rm -f /etc/kubernetes/pki/etcd/server*
...
 then
-  kubectl -n $NAMESPACE create secret generic $FULLNAME-etcd-cert \
-    --from-file=/etc/kubernetes/pki/etcd/ca.crt \
-    --from-file=/etc/kubernetes/pki/etcd/server.crt \
-    --from-file=/etc/kubernetes/pki/etcd/server.key
-  kubectl -n $NAMESPACE label secret $FULLNAME-etcd-cert $LABELS
+  kubectl -n $NAMESPACE delete secret $FULLNAME-etcd-cert
 fi
-if secret_not_found $FULLNAME-apiserver-cert
+kubectl -n $NAMESPACE create secret generic $FULLNAME-etcd-cert \
```
kd7lxl, line 49: It might be better to use `kubectl patch` here instead of deleting.

PR #322, opened 2022-07-14
fix CI ingress-nginx installation

kd7lxl:
The file we relied on previously to detect the latest ingress-nginx version was removed, so I changed it to use the release tags.

https://github.com/kubernetes/ingress-nginx/pull/8778

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -236,7 +236,7 @@ $(CHART_READMES):
 
 chart-readmes: $(CHART_READMES)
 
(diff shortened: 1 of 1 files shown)


kd7lxl:
Note that the `upgrade` job is expected to fail because it tries to install based on what is currently in `main`, which is broken, and the reason for this PR.

hramazani (approved):
LGTM

PR #379, opened 2023-01-26
migrate to newer kubeadm spec for k8s v1.24+

kd7lxl:


kd7lxl:
This may be useful someday, but the problem with the CI is unrelated.

Issue #207, opened 2021-05-06
konk-service resources should be deleted upon KonkService deletion


kd7lxl:
The resources created by konk-service are not managed by the helm-operator, nor do they have ownerReferences. This leaves nothing to clean them up after a deletion. After deleting a `KonkService`, these resources remain:
https://github.com/infobloxopen/konk/blob/a63fd294848fcaacac3ef0b64449d52561a564a5/helm-charts/konk-service/templates/apiservice-configmap.yaml#L39-L83

Some mechanism should be used to cleanup these resources, such as refactoring so the helm-operator manages them, or adding `ownerReferences` so they are cleaned up by kube-controller-manager.

PR #437, opened 2023-07-25
Added pod def to delete resources in the konk when the corresponding KonkService is deleted

abalaven:
Should resolve https://github.com/infobloxopen/konk/issues/207

### DIFF
--- a/helm-charts/konk-service/templates/apiservice-configmap.yaml
+++ b/helm-charts/konk-service/templates/apiservice-configmap.yaml
@@ -30,6 +30,24 @@ data:
     fi
     echo 'Done' > /tmp/healthy
 
+  delete-api-service.sh: |
+    #!/bin/bash
+    set -xe
+
+    CERT=$(cat certs/ca.crt | base64)
+    CERT=${CERT//$'\n'/}
+    cat /mounts/api-service.yaml.in | sed "s/{{` {{ SERVICENAME }} `}}/${SERVICENAME}/g" | sed "s/{{` {{ NAMESPACE }} `}}/${NAMESPACE}/g" | sed "s/{{` {{ CERT }} `}}/${CERT}/g" > /gen/api-service.yaml
+    delay=2
+    until kubectl delete -f /gen/api-service.yaml
+    do
+      sleep $delay
+      let delay="delay*2"
+      if [ "$delay" -gt "60" ]
+      then
+        delay=60
+      fi
+    done
+
   api-service.yaml.in: |
     apiVersion: v1
     kind: Namespace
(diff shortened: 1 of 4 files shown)


helm-charts/konk-service/templates/delete-apiservice-deployment.yaml
```diff
@@ -0,0 +1,71 @@
...
+    imagePullPolicy: {{ .Values.kind.image.pullPolicy }}
+    command: ["/bin/bash", "-c"]
+    args:
+      - |
+        while true
+        do
+          /mounts/delete-api-service.sh
+          date
+          sleep 30
+        done
```
kd7lxl, line 26: Don't you want this to just run once and exit?

helm-charts/konk-service/templates/delete-apiservice-deployment.yaml
```diff
@@ -0,0 +1,71 @@
+apiVersion: v1
+kind: Pod
```
kd7lxl, line 2: Would this be better as a Job?

abalaven: It shouldn't matter, right? Whether it's a job/pod, it will get deleted along with the rest of the helm-chart when the konk-service is deleted.

kd7lxl: It would change the retry behavior, but this implementation doesn't depend on that.

kd7lxl (approved):

helm-charts/konk-service/templates/delete-apiservice-deployment.yaml
```diff
@@ -0,0 +1,66 @@
...
+    - name: mounts
+      mountPath: "mounts"
+      readOnly: true
+    - name: gen
+      mountPath: "gen"
+    readinessProbe:
+      exec:
+        command:
+        - cat
+        - /tmp/healthy
```
kd7lxl, line 47: Looks like it never finds this file, so never becomes ready. I don't see any code that would create it.

helm-charts/konk-service/templates/apiservice-configmap.yaml
```diff
@@ -30,6 +30,25 @@ data:
...
+    delay=2
+    until kubectl delete -f /gen/api-service.yaml
+    do
+      sleep $delay
+      let delay="delay*2"
+      if [ "$delay" -gt "60" ]
+      then
+        delay=60
+      fi
+    done
```
kd7lxl, line 49: You might not want to retry here. If it fails, it could be that the konk has been deleted, in which case there's no longer a need to delete the API service. Or maybe the APIService has already been deleted by something else.

abalaven: Removing the retry and converting the pod to a job.

kd7lxl (approved):

PR #455, opened 2023-09-01
check etcd in konk readiness

kd7lxl:
Previously we were only testing the konk apiserver for liveness. This changes it to use the readiness check, which checks a number of things including the etcd connection. See https://kubernetes.io/docs/reference/using-api/health-checks/#api-endpoints-for-health

Also cleaned up a bunch of things in the CI to support testing this.


drewwells (approved):
Fancy, thanks

PR #471, opened 2023-10-26
Truncate the delete apiservice job name

abalaven:
Observed an error where a long konk-service release name could potentially prevent a delete-apiservice job from triggering as it would surpass the 63 character limit allowed for job names. This PR resolves this by truncating the length of the job name to less than 63 characters.

### DIFF
--- a/helm-charts/konk-service/templates/delete-apiservice-deployment.yaml
+++ b/helm-charts/konk-service/templates/delete-apiservice-deployment.yaml
@@ -1,7 +1,7 @@
 apiVersion: batch/v1
 kind: Job
 metadata:
-  name: {{ include "konk-service.fullname" . }}-delete-apiservice
+  name: {{ include "konk-service.fullname" . | trunc 43 }}-delete-apiservice
   namespace: {{ .Release.Namespace }}
   labels:
     {{- include "konk-service.labels" . | nindent 4 }}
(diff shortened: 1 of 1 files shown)


helm-charts/konk-service/templates/delete-apiservice-deployment.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: batch/v1
 kind: Job
 metadata:
-  name: {{ include "konk-service.fullname" . }}-delete-apiservice
+  name: {{ include "konk-service.fullname" . | trunc 43 }}-delete-apiservice
```
kd7lxl, line 4: This is fine, but it might be a little more obvious truncate the whole string to the actual limit like
```suggestion
  name: {{ printf "%s-delete-apiservice (include "konk-service.fullname" .) | trunc 63 }}
```

kd7lxl (approved):

helm-charts/konk-service/templates/delete-apiservice-deployment.yaml
kd7lxl: Never mind, I think I prefer the way you did it originally. A really long name that truncates the entire `-delete-apiservice` part off the end could result in an ambiguous name.

kd7lxl: One thing I would do either way is trim double `-` like the helpers do:
```suggestion
  name: {{ include "konk-service.fullname" . | trunc 43  | trimSuffix "-" }}-delete-apiservice
```

kd7lxl (approved):

kd7lxl (approved):

PR #593, opened 2026-03-23
fix: push new distroless images in CI and load correct tags in operator

rsatal:
- Jenkinsfile: add docker-build/push targets for konk-app, konk-provision, and konk-service — these images were never pushed by CI after being introduced in #582
- Makefile: simplify image-tag-values.yaml generation (remove kubeadm dependency, add kind.image.tag); make docker-build depend on it so tags are always regenerated before baking into the operator image
- watches.yaml: add valuesFiles for image-tag-values.yaml so operator loads correct image tags at runtime instead of falling back to appVersion
- image-tag-values.yaml: update bootstrap tags for current build

### DIFF
--- a/helm-charts/konk/image-tag-values.yaml
+++ b/helm-charts/konk/image-tag-values.yaml
@@ -1,13 +1,13 @@
-# kubernetes v1.25.8
+# kubernetes $(K8S_RELEASE)
 apiserver:
   image:
-    tag: v1.25.8-g9fefaa7
+    tag: $(K8S_RELEASE)-$(GIT_SHORT)
 etcd:
   image:
-    tag: v3.6.8
+    tag: $(ETCD_VERSION)
 provision:
   image:
-    tag: v0.2.1-150-g9fefaa7-j11
+    tag: $(GIT_VERSION)
 kind:
   image:
-    tag: v0.2.1-150-g9fefaa7-j11
+    tag: $(GIT_VERSION)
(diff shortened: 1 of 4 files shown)


helm-charts/konk/image-tag-values.yaml
```diff
@@ -1,7 +1,13 @@
 # kubernetes v1.25.8
 apiserver:
   image:
-    tag: v1.25.8
+    tag: v1.25.8-g9fefaa7
 etcd:
   image:
-    tag: 3.4.9-1
+    tag: v3.6.8
```
daniel-garcia: these should be templated out

helm-charts/konk/image-tag-values.yaml
```diff
@@ -1,7 +1,13 @@
...
+provision:
+  image:
+    tag: v0.2.1-150-g9fefaa7-j11
```
daniel-garcia: these should be templated out

daniel-garcia (changes requested):

helm-charts/konk/image-tag-values.yaml
rsatal: Addressed

helm-charts/konk/image-tag-values.yaml
rsatal: Addressed

daniel-garcia (approved):

vbuddhavarapu-infoblox (approved):

PR #659, opened 2026-07-10
Log actual Helm error on upgrade failure

rsatal:
## Summary
- Operator-sdk's `UpgradeRelease()` drops the real Helm upgrade error (hook timeout, resource conflict, etc.) and replaces it with a bare sentinel `"upgrade failed; rollback required"`
- The actual error was printed to stdout via `fmt.Printf` — invisible in structured JSON logs
- This patch wraps the underlying Helm error into `ErrUpgradeFailed` via `fmt.Errorf("%w: %v", ErrUpgradeFailed, err)` so the real cause propagates into operator logs and CR status conditions

## Before
```
{"level":"error","msg":"Release failed","error":"upgrade failed; rollback required"}
```

## After
```
{"level":"error","msg":"Release failed","error":"upgrade failed; rollback required: pre-upgrade hook \"bulk-konk-etcd-fix-helm-orphans\" failed: job failed: BackoffLimitExceeded"}
```

## Context
Discovered while debugging etcd upgrade failure on us-dev-4 where the upgrade error was completely opaque — required manual correlation of events, Jobs, and webhook logs across namespaces to find the root cause (VPC admission webhook blocking hook pods).

## Test plan
- [ ] Docker build succeeds with patch applied
- [ ] Deploy to dev cluster and trigger an upgrade — verify operator logs include the underlying Helm error
- [ ] Verify `errors.Is(err, ErrUpgradeFailed)` still works (rollback path still triggers)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,8 +6,10 @@ RUN apk add --no-cache git make bash
 
 # Clone and build operator-sdk/helm-operator
 WORKDIR /workspace
+COPY patches/ /workspace/patches/
 RUN git clone --depth 1 --branch v1.42.0 https://github.com/operator-framework/operator-sdk.git && \
     cd operator-sdk && \
+    git apply /workspace/patches/operator-sdk-upgrade-error-logging.patch && \
     go get golang.org/x/crypto@v0.52.0 \
            golang.org/x/net@v0.55.0 \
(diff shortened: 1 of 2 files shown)


rsatal:
Pushed directly to release/cve-remediations-july26 instead

PR #674, opened 2026-08-03
feat(konk-operator): add registry mirror support

sreejith-infoblox:
## Summary

Pulling this chart's images from anywhere other than public GHCR previously meant overriding four separate values — `image.repository` plus three `relatedImages.*Repository` entries — in every deployment. This adds a single `registry` block that covers all of them:

```yaml
registry:
  host: ""   # e.g. registry.example.com
  org: ""    # optional path segment
```

When `host` is set, each image resolves to `{host}/{org}/{name}`, where `{name}` is the last path segment of that image's own `repository`. The `org` segment is dropped when empty, giving a flat `{host}/{name}` layout. With no host set the chart pulls the public images exactly as before, so this is opt-in.

`registry.host` falls back to `global.ib0.services.registry.host` when unset, letting a deployment platform supply the mirror host without any per-cluster configuration.

## Other changes

- `podAnnotations` now defaults to Prometheus scrape annotations for the operator's metrics endpoint (`:8080`, the helm-operator default). Set to `{}` to opt out.
- `crds.create` now defaults to `true`. The template renders whatever is in the chart's `crds/` directory, which `make package` populates from `config/crd/bases`, so a source-tree render is unaffected.
- Grouped each `relatedImages` tag beside its repository, and corrected comments that described an appVersion fallback the template does not implement.

## Test plan

- [x] `helm lint` passes
- [x] Rendered output is unchanged for existing deployments, except where a platform registry host is present — those now resolve to the mirror, which is the intended behaviour change
- [x] Mirror resolution verified for three cases: no mirror (public GHCR), host only (flat layout), and host + org
- [x] `helm template` for a fresh deployment renders the expected manifests

### DIFF
--- a/helm-charts/konk-operator/templates/_helpers.tpl
+++ b/helm-charts/konk-operator/templates/_helpers.tpl
@@ -61,3 +61,24 @@ Create the name of the service account to use
 {{- default "default" .Values.serviceAccount.name }}
 {{- end }}
 {{- end }}
+
+
+{{/*
+Resolve the repository for one image.
+
+The mirror host is registry.host, or the ib0 platform registry when that is
+empty. Given one, returns {host}/{org}/{name} — {name} being the last path
+segment of the image's own repository, and the {org} segment dropped when
+registry.org is empty. Given neither, returns the repository unchanged.
+
+Usage: include "konk-operator.imageRepo" (dict "ctx" . "repo" .Values.image.repository)
+*/}}
+{{- define "konk-operator.imageRepo" -}}
+{{- $registry := .ctx.Values.registry -}}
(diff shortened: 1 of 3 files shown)


sreejith-infoblox:
Dropped the `crds.create: false` → `true` change from this PR (force-pushed).

It broke the Makefile install path CI uses: `make deploy-crds` applies the CRDs outside Helm by design, and with the flag defaulting to `true` the packaged chart also renders them into the release, so Helm fails trying to adopt resources it doesn't own. Details and the same fix for `release/ib0-test` are in #677.

A source-tree render can't catch this — `crds.yaml` globs `crds/*.yaml`, which is empty until `make package` populates it from `config/crd/bases`, so `helm template`/`helm lint` render zero CRDs either way.

PR #696, opened 2026-09-09
zero-config: migrate konk-operator chart to ib0

sreejith-infoblox:
Makes the konk-operator chart configurable for any registry, fixes two bugs the
chart carried, and removes a double-scrape trap. Four commits.

### 1. Registry mirror support (`7a8e9e6`)

Pulling this chart's images from anywhere other than public GHCR previously meant
overriding four separate values in every deployment. A single `registry` block now
redirects all of them:

```yaml
registry:
  host: ""              # empty -> falls back to global.ib0.services.registry.host
  org: infobloxcto
```

Each image resolves to `{host}/{org}/{name}`, where `{name}` is the last path segment
of that image's own `repository`. With no host anywhere, the public `repository`
defaults are used unchanged.

**This is not ib0-only.** `registry.host` is an ordinary chart value — any deployment
can set it. What is ib0-specific is only the *automatic* part: the fallback to
`global.ib0.services.registry.host`, which the ib0 platform injects, so ib0 clusters
mirror with no configuration at all. Non-ib0 clusters set `registry.host` explicitly
and need no chart change. The EU proxy layout is expressible too, since `org`
tolerates slashes:

| values | resolves to |
|---|---|
| *(defaults, no ib0)* | `ghcr.io/infobloxopen/konk-app` |
| ib0 cluster, `host: ""` | `harbor.services.sdp.infoblox.com/infobloxcto/konk-app` |
| `host: harbor.services.sdp.infoblox.com` | `harbor.services.sdp.infoblox.com/infobloxcto/konk-app` |
| `host: harbor.services.eu.sdp.infoblox.com`, `org: harbor-proxy-cache/infobloxcto` | `harbor.services.eu.sdp.infoblox.com/harbor-proxy-cache/infobloxcto/konk-app` |

#### How far the ib0 fallback actually reaches today

Worth knowing before anyone relies on the automatic path: in the DC repo the
`global.ib0` descriptor comes from `envs/box-dev/us-dev-5/zero-config-values.yaml`,
whose `global:` block is merged into every app's values on that cluster. That is the
**only** `zero-config-values.yaml` in the repo, and the file marks itself
`TRANSITIONAL: until webhook injection is available`, with
`services.registry.host` contributed by the platform team.

So the zero-config path currently serves exactly one cluster:

| cluster | `global.ib0` reaches konk-operator? | `services.registry` |
|---|---|---|
| us-dev-5 | yes | `host: harbor.services.sdp.infoblox.com` |
| us-dev-2, us-dev-4, us-stg-1, eu-stg-1 | no — an `ib0` block exists only inside unrelated app values (`karpenter-values.yaml`, `atlas-cubejs-async-values.yaml`) | absent, `services` empty |
| us-com-1, eu-com-1, prd-1, gov-stg-2, gov-prd-2, gstg-1 | no `ib0` at all | absent |

Every other cluster therefore sets `registry.host` explicitly — which is one line and
renders identically. Do **not** hand-author a `zero-config-values.yaml` to get the
fallback: that file carries platform-owned cluster facts (AWS account and subnet IDs,
CA ARNs, EIP allocations), and the admission webhook is meant to supply it fleet-wide
later.

Explicit values stay forward-compatible, since resolution is
`registry.host | default global.ib0.services.registry.host` — when the webhook lands,
an explicit host keeps winning and can be deleted then. That precedence is also the one
trap: if a cluster is later injected with a *different* registry, a stale explicit host
silently overrides it. Gov is the realistic case, where the platform registry
(`harbor-services.sdp.prd.infoblox-fedcloud.com`) is not where konk images live
(`harbor.services.sdp.infoblox.com`). Prefer setting `registry.host` at the lifecycle
level, where it is easy to find and remove, over scattering it per cluster.

### 2. `crds.create` restored to `false` (`704dfec`)

#675 had promoted it to `true` because every deployment sets it, but that broke the
Makefile install path CI uses: `make deploy-crds` applies the CRDs without Helm
ownership metadata (deliberately — *"There is no support at this time for upgrading or
deleting CRDs using Helm"*), so a chart that also renders them into the release makes
Helm try to adopt resources it does not own. Note a source-tree render cannot catch
this: `crds.yaml` globs `crds/*.yaml`, which is empty until `make package` populates it.

### 3. `relatedImages` tags fall back to `appVersion` (`43cfbab`)

values.yaml documented these tags as overriding "the chart's default appVersion
fallback", but the template had no such fallback — an unset tag rendered a bare
`repo:` with an empty tag, which for the `fix-helm-orphans` init container is an
invalid image reference, so the pod never started.

`make package` builds with `--app-version ${GIT_VERSION}` (chart `--version` gets the
extra `-jNN`), so `.Chart.AppVersion` is already exactly the tag the images were built
with. The fallback is therefore self-consistent by construction: chart *vX* always
ships alongside images tagged *vX* from the same build.

This also replaces a fallback that was already broken. Previously an unset tag fell
through to the **konk subchart's** appVersion, which is `v1.25.8`:

```
ghcr.io/infobloxopen/konk-app:v1.25.8              -> HTTP 404   (old fallback)
ghcr.io/infobloxopen/konk-app:v0.2.1-164-gbd3f28a  -> HTTP 200   (new fallback)
```

### 4. Scrape annotations gated on the PodMonitor (`0ac5334`)

#675 also promoted the `prometheus.io` scrape annotations into `podAnnotations`.
Correct in intent — every deployment was setting them by hand — but it put them on an
ungated passthrough, and combined with the PodMonitor from #681/#699 a pod can carry
both discovery paths at once. Alloy's annotation discovery and its
`prometheus.operator.podmonitors` component both run with no selector, so such a pod is
scraped twice and every series is duplicated, the copies differing only in
`job`/`instance`.

The etcd chart already solved this in #689/#699. konk-operator had no equivalent, which
is why us-dev-2 and us-dev-5 carry a hand-written `podAnnotations: null` in the DC repo
purely to de-duplicate. The scrape pair moves to `metrics.podAnnotations` behind the
same gate; `podAnnotations` returns to `{}` for genuine user annotations.

This partially reverts #675's `podAnnotations` default — annotation scraping is now
requested via `metrics.enabled` instead.

## Verification

`helm lint` clean. Renders with `--api-versions monitoring.coreos.com/v1`:

| scenario | annotations | PodMonitor |
|---|---|---|
| chart defaults | 0 | 0 |
| `metrics.enabled` | 2 | 0 |
| `metrics.enabled` + `podMonitor.enabled` | **0** | 1 |
| DC `envs/konk-operator-values.yaml` | 2 | 0 — unchanged |
| DC us-dev-5 (`null` + podMonitor) | 0 | 1 — unchanged |

Rendered against every deployed cluster's real DC values: **no cluster changes
behaviour on merge.** `registry.host` defaults to `""`, and only us-dev-5 has
`global.ib0` in its values, so every other cluster keeps its current repositories
verbatim. Confirmed byte-identical for gov-prd-2: today's 8 pinned lines and a single
`registry.host` produce the same manifest.

## Expected changes in deployment-configurations

Nothing is required at merge time — the chart is inert until DC opts in. When you do:

**One line, fleet-wide,** in `envs/konk-operator-values.yaml`:

```yaml
registry:
  host: harbor.services.sdp.infoblox.com
```

That host is simultaneously what ib0 injects into us-dev-5 and what gov-prd-2,
gov-stg-2 and us-dev-4 hardcode today.

**Then delete the now-redundant overrides.** Only three files fleet-wide carry them —
`box-dev/us-dev-4`, `gov-box-prod/gov-prd-2`, `gov-box-stage/gov-stg-2`. Each collapses
from this:

```yaml
image:
  repository: harbor.services.sdp.infoblox.com/infobloxcto/konk
relatedImages:
  apiserver: v0.2.1-168-gb8dabd8
  provision: v0.2.1-168-gb8dabd8
  kind: v0.2.1-168-gb8dabd8
  apiserverRepository: harbor.services.sdp.infoblox.com/infobloxcto/konk-app
  provisionRepository: harbor.services.sdp.infoblox.com/infobloxcto/konk-provision
  kindRepository: harbor.services.sdp.infoblox.com/infobloxcto/konk-service
```

…to nothing. Image refs then follow `konk-operator-version.txt` on their own. All three
clusters already pin tags that equal their chart's appVersion exactly, so this is a pure
refactor:

| cluster | pinned chart | appVersion | tags set today |
|---|---|---|---|
| us-dev-4 | `v0.2.1-168-gb8dabd8-j207` | `v0.2.1-168-gb8dabd8` | same ×3 |
| gov-prd-2 | `v0.2.1-164-gbd3f28a-j203` | `v0.2.1-164-gbd3f28a` | same ×3 |
| gov-stg-2 | `v0.2.1-164-gbd3f28a-j203` | `v0.2.1-164-gbd3f28a` | same ×3 |

**Why this matters beyond tidiness:** bumping only `konk-operator-version.txt` while
keeping the pinned tags produces a split-version deployment — the operator moves to the
new build while it reconciles every Konk/KonkService CR with the *previous* apiserver,
provision and konk-service images. That is what the old file's *"Must be updated when
konk chart version changes"* comment guarded against; it becomes structurally impossible.

**For commit 4,** replace the global `podAnnotations` block with `metrics.enabled: true`,
then drop `podAnnotations: null` from `box-dev/us-dev-2` and `box-dev/us-dev-5`.

### Rollout cautions

- **The global host is also a GHCR -> Harbor migration** for the 8 clusters currently on
  the chart's public defaults (us-dev-2, box-3, us-com-1, eu-com-1, us-stg-1, eu-stg-1,
  gstg-1, prd-1). Confirm `infobloxcto/konk{,-app,-provision,-service}:<appVersion>`
  exists in Harbor for each cluster's pinned version first — `konk-app:v0.2.1-138-g8b64bf7`
  already 404s in GHCR, so old pins are genuinely at risk. To decouple, set
  `registry.host` per lifecycle instead of at the top level.
- **`envs/com-prod/us-com-1/konk-operator-version.txt` is a symlink** to
  `prod/prd-1/konk-operator-version.txt` — those two clusters must move together.
- **`integration` is on `v0.2.1-189-g6fc79bc-j39`**, a later commit than the current
  release line; bumping it backwards would be a rollback.

## Known gap

`konk-operator.imageRepo` dereferences `.Values.global.ib0.services.registry.host`
unguarded, so `--set global.ib0=null` fails with `nil pointer evaluating
interface {}.services`. No current consumer does that — Helm's deep merge preserves the
chart's own `global.ib0` default even when a parent sets other `global` keys — but it is
worth a guard once DC values depend on this block. Not included here to keep the diff
reviewable.


rsatal:
@sreejith-infoblox — pushed one commit to this branch via maintainer-edit (`0ac5334`), heads-up since it's your PR. Happy to move it to a separate PR if you'd rather keep this one focused.

**Why:** #675 promoted the `prometheus.io` scrape annotations into `podAnnotations`, which is correct in intent — every deployment was setting them by hand — but it landed them on an ungated passthrough. Combined with the PodMonitor that came from #681/#699, a pod can now carry both discovery paths at once. Alloy's annotation discovery and its `prometheus.operator.podmonitors` component both run with no selector, so such a pod is scraped twice and every series is duplicated, the copies differing only in `job`/`instance`.

The etcd chart already solved this in #689/#699 by emitting its annotations only when the PodMonitor isn't doing the job. konk-operator had no equivalent, which is why us-dev-2 and us-dev-5 carry a hand-written `podAnnotations: null` in the DC repo purely to de-duplicate.

**What changed:** scrape annotations move to `metrics.podAnnotations` behind the same gate etcd uses; `podAnnotations` returns to `{}` for genuine user annotations. This partially reverts the `podAnnotations` default from #675 — annotation scraping is now requested via `metrics.enabled` instead.

**Verified by render** (`--api-versions monitoring.coreos.com/v1`), `helm lint` clean:

| scenario | annotations | PodMonitor |
|---|---|---|
| chart defaults | 0 | 0 |
| `metrics.enabled` | 2 | 0 |
| `metrics.enabled` + `podMonitor.enabled` | **0** | 1 |
| DC `envs/konk-operator-values.yaml` | 2 | 0 — unchanged |
| DC us-dev-5 (`null` + podMonitor) | 0 | 1 — unchanged |

Both DC paths render byte-identical, so no cluster changes behaviour on merge. Follow-up on the DC side: swap the global `podAnnotations` for `metrics.enabled: true` and drop the per-cluster `null` overrides.

vijaylakshmi-subramanian (approved):

PR #683, opened 2026-08-14
fix(etcd): prevent Karpenter eviction loops (release/upgrade-etcd)

rsatal:
## Summary

Port of [PR #682](https://github.com/infobloxopen/konk/pull/682) to the `release/upgrade-etcd` branch.

Fixes the etcd Karpenter consolidation eviction loop on us-dev-2 (DEVOPS-47739). The `WhenEmptyOrUnderutilized` Karpenter consolidation policy was evicting nodes hosting etcd pods because their CPU requests (10m) were far below actual usage, making nodes appear underutilized. Each eviction triggered a 1–2 min EBS volume detach/reattach cycle, cascading into API downtime for all konk-served services.

Re-implemented from scratch (not cherry-picked) due to structural differences between `main` and `release/upgrade-etcd`.

---

### `helm-charts/etcd` (1.1.2 → 1.2.2)

**`values.yaml`**
- Added `pdb` section (`enabled: true`, `minAvailable: 2`)
- Populated `tolerations` with `infoblox.com/do-not-disrupt: NoSchedule` toleration (harmless on clusters without the taint)
- Populated `affinity` with soft (`preferredDuringSchedulingIgnoredDuringExecution`) node affinity for `node-group-type: stable` — falls back to any node on clusters without `stable-node-pool`

**`templates/poddisruptionbudget.yaml`** _(new)_
- Creates a `PodDisruptionBudget` when `pdb.enabled=true` and `statefulset.replicaCount > 1`
- `minAvailable: 2` on a 3-replica cluster → `disruptionsAllowed: 1`, preventing simultaneous eviction of multiple etcd members while still allowing rolling node drains
- Guard on `replicaCount > 1` prevents single-replica dev deployments from getting a PDB that blocks all eviction

### `helm-charts/konk` (0.1.0 → 0.2.0)

**`values.yaml`** — critical fix
- `etcd.resources.requests.cpu`: `10m → 200m` — **this is the actual source of the 10m**; the konk operator passes `.Values.etcd` as helm values into the Etcd CR, overriding etcd chart defaults
- `etcd.resources.requests.memory`: `64Mi → 128Mi` — aligns with actual etcd memory footprint
- Added `etcd.tolerations` — tolerate `infoblox.com/do-not-disrupt: NoSchedule` taint on stable-node-pool nodes
- Added `etcd.affinity` — soft preference for `node-group-type: stable` nodes; no-op on clusters without the pool

### `helm-charts/konk-service` (0.1.0 → 0.2.0)

[...]

PR #699, opened 2026-09-10
Port etcd 3.7.1 CVE fix, konk dashboards and claimName default from release/upgrade-etcd

rsatal:
## Summary

Brings the still-wanted work from `release/upgrade-etcd` onto a branch cut fresh from
`main`, plus fixes found while reviewing it. The source range is
`8b64bf7..release/upgrade-etcd` (25 commits), but auditing it against `main` first
showed **9 of the 25 were already on `main`** and most of the remainder were
superseded, redundant, or unsafe to replay.

Headline changes:

- **etcd `3.7.0` → `3.7.1`**, fixing GHSA-hrxh-6v49-42gf (HIGH, grpc in the
  etcd/etcdctl/etcdutl binaries). `main` was still on `3.7.0`.
- **`persistence.claimName` now defaults to `data-v2`** — read the deployment note
  below before rolling this out.
- **New konk-services Grafana dashboard** plus large expansions to the etcd and
  konk-operator dashboards, and metrics scraping for the operator itself.
- **Operator Deployment now uses the `Recreate` strategy**, closing a window where two
  operators reconciled the same CRs concurrently and corrupted Helm release history.

## ⚠️ Deployment note: `claimName` default change

`persistence.claimName` defaults to `data-v2` instead of `data`. A StatefulSet's
`volumeClaimTemplate` name is **immutable**, so ordering matters:

- For deployments already overriding `claimName: data-v2` in the DC repo — the large
  majority, including both com-prod clusters — this is a **no-op**. Same value, no
  StatefulSet change.
- For any deployment still on `data`, this chart must **not** arrive before that
  deployment has been migrated. Otherwise Helm hands the API server a `data` →
  `data-v2` change, the API server rejects it, and the helm-operator loops in
  failed-upgrade/rollback.

The pre-upgrade hook that automated the StatefulSet recreate on `release/upgrade-etcd`
is deliberately **not** in this chart line (see Excluded), so recovery would be a manual
`kubectl delete sts <name> --cascade=orphan`. The remaining deployments are being
migrated first using the existing hook-bearing operator build, after which this lands as
a no-op everywhere.

## Included

| Commit | Source | What |
|---|---|---|
| `f448d9b` | #631 | Parameterizes the etcd `volumeClaimTemplate` name via `persistence.claimName`. |
| `7190e58` | `9deaa9d` | etcd image `3.7.0` → `3.7.1` — GHSA-hrxh-6v49-42gf. |
| `b95ebb0` | #688 #689 #690 #691 | Grafana dashboards + metrics scraping. |
| `843ddc6` | — | Makes konk-service pod scheduling preferences opt-in. |
| `97e1360` | — | Removes a hardcoded namespace list from the konk-services dashboard. |
| `4c1bf77` | — | Operator Deployment uses `Recreate` so two operators never overlap. |
| `14243f7` | — | Defaults `persistence.claimName` to `data-v2`. |
| `dc4d4a4` | CI | Chart README regeneration (`readme` job). |
| `1b06f5d` + `2438709` | — | Instance-label templating, then reverted — **net zero**. |

### `7190e58` — reconciliation notes

Applied by hand rather than as a straight cherry-pick, because the upstream commit was
written against an older tree:

- Kept `main`'s parameterized `ETCD_IMG ?= gcr.io/etcd-development/etcd:$(ETCD_VERSION)`
  and bumped `ETCD_VERSION` to `v3.7.1`, rather than taking the upstream hardcoded form.
- Chart version continues upward rather than regressing to the upstream commit's `1.1.1`.

### `b95ebb0` — observability

The four dashboard commits all pile onto the same three JSON files, so this takes their
**final state per file in one pass** instead of replaying four commits and resolving the
same conflicts four times. Net result is identical.

- `etcd.json` 12 → 17 panels, `konk-operator.json` 5 → 27 panels, and a new
  `konk-services.json` (38 panels) with its ConfigMap template.
- Verified the branch's dashboards are a **strict superset** of `main`'s — nothing is
  lost. The one apparently `main`-only panel, `Proposals Failed / Pending`, was *split*
  into separate Failed and Pending panels plus a new Committed Rate panel; both original
  metrics are retained.
- #684's Grafana 10 work survives: `dashboard.yaml` is byte-identical to `main`'s, and
  both dashboards are all-`timeseries` with zero deprecated `graph` panels.
- konk-operator gains a `metrics` values block, a named `metrics` container port, and a
  PodMonitor for its own `/metrics` endpoint.
- The etcd PodMonitor is now gated on the `monitoring.coreos.com/v1` capability, so
  enabling metrics cannot fail the install on a cluster with no prometheus-operator.
- etcd emits the `prometheus.io` scrape annotations **only** when the PodMonitor is off,
  so collectors honouring both mechanisms stop scraping each pod twice and doubling every
  series.

`konk-operator/values.yaml` and `deployment.yaml` were edited surgically rather than
copied wholesale: the upstream branch predates `main`'s `fix-helm-orphans` init
container, its `relatedImages` plumbing, and the move to the GHCR image repo.

### `97e1360` — dashboard portability

`konk-services.json` baked a snapshot of the commercial clusters' tenant namespaces into
the JSON in **five** places. Since this ships inside the operator image there is no value
to override it per cluster, so any cluster with a different namespace set reported wrong
numbers. Three distinct problems:

- `Backends with endpoints` (2 queries) ignored `$namespace` entirely, and with
  `or vector(0)` rendered a confident `0` rather than "No data" where none of the listed
  namespaces exist — a wrong answer that looks healthy. Now scoped by `$namespace`.
- `Not-ready pods per namespace` carried both `$namespace` and the hardcoded matcher in
  one selector. PromQL ANDs repeated matchers on a label, so the effective selection was
  the intersection: picking a namespace outside the list returned nothing while the
  dropdown appeared to work.
- The `$namespace` variable's own `allValue` **was** the hardcoded list, so selecting
  "All" capped the whole dashboard, including the 24 panels that correctly use
  `$namespace`.

The list was redundant as well as unportable: the variable is
`label_values(kube_pod_status_ready{pod=~".*konk-service.*"}, namespace)`, so it already
resolves to exactly the namespaces running konk-service.

### `4c1bf77` — operator rollout

Not part of the etcd port. It is here because the `End To End (true, v1.25.11, main)`
leg was failing intermittently on this branch — three attempts on an unchanged commit
(fail, fail, pass) — and the cause turned out to be a pre-existing race on `main` that
this PR triggers far more often than most changes do.

The operator Deployment had `replicas: 1` and **no `strategy:` block**, so Kubernetes
defaults applied: `maxUnavailable: 25%` → floor → **0**, `maxSurge: 25%` → ceil → **1**.
Because one pod must stay Available, the new pod had to become Ready *before* the old was
removed — so **every** operator upgrade guaranteed a window with two operator pods
running. Nothing arbitrates between them: the chart ships a leader-election
`RoleBinding`, but `--leader-elect` is never passed to the container and the `Role` it
binds to exists only in `config/rbac/`, not in this chart. Both instances therefore
reconcile the same `Konk` and `Etcd` CRs and drive Helm releases concurrently, corrupting
each other's release history:

```
failed to update: secrets "sh.helm.release.v1.runner-konk-etcd.v3" not found
failed to install release: release: already exists
rollback failed: release has no 1 version
```

Once the history is gone the operator cannot converge, so etcd never becomes healthy and
the konk apiserver never passes `readyz`.

It is load-correlated rather than random, which is why it has not bitten before: #681
touched only the etcd chart, so one release was in flight during the overlap, and it
passed 3/3. This PR touches the `konk`, `etcd` and `konk-service` charts — three releases
in the same window.

`Recreate` also repairs a second problem. The `fix-helm-orphans` init container states it
must "start with a clean slate on every operator restart", and it deletes non-deployed
release secrets — but `pending-upgrade` is indistinguishable from an upgrade the outgoing
operator is running *right now*. That premise was false by construction; it is now true.

**Trade-off, stated plainly:** with `Recreate`, if the new pod never becomes Ready you
have *zero* operators, where `RollingUpdate` would have kept the old one serving. Judged
worth it because:

- `RollingUpdate`'s safety here was partly illusory — the init container runs to
  completion *before* readiness is evaluated, so even a rollout that ultimately failed
  had already deleted the live operator's pending release secrets. It protected against
  downtime, not against the damage.
- It is a control-plane outage only. Existing konk clusters, etcd and apiservers keep
  serving; what stops is reconciliation.
- It is visible and reversible: `/healthz` and `/readyz` probes exist, deploys use
  `helm upgrade --wait` so a bad rollout fails loudly, and recovery is `helm rollback`.

No strategy offers both — avoiding overlap *means* killing the old pod first. The only
mechanism that would is leader election, which needs the missing `Role` and still would
not gate the init container, since that runs before any lock is acquired.

## Excluded

| Commit(s) | Why |
|---|---|
| `f8540d7` | ⛔ **Would cause an outage.** Flips the etcd chart to read `.Values.statefulset.replicaCount` but does not touch `konk/values.yaml`, which still sends `replicaCount`. `main` is consistent on the flat key end-to-end and renders 3 replicas correctly. Applied alone this drops every konk-deployed etcd from 3 replicas to 1 — the exact failure `2606165`'s own commit body describes: *"causing the StatefulSet to scale from 3 to 1 on upgrade and leaving etcd-0 with stale 3-node cluster data that failed health checks."* |
| `2606165` (#683) | Fully redundant — self-described as a port of `main`'s #682, and `main` has the original plus everything since. The PDB template, `pdb` values, etcd CPU/memory requests and the CI timeout are all already present (the timeout is already more generous). Taking it would also regress konk-service, which is further ahead on `main`. |
| `1de007e` (#635) | Superseded by `main`'s own `89ed1cf` (#637) — the same fix re-landed 4 days later. The `helmAnnotations` helper and its ~33 template usages are already present. |
| `e13f6ed` | Superseded — `main` already moved to `cgr.dev/infoblox.com/etcd:3.7.0` via `b4dd9fa`, three weeks earlier. |
| `fd9ed6b` (#634), `372db4e` (#636) | The StatefulSet recreate hook. Migration-only: both touch nothing but `recreate-statefulset-hook.yaml` and its values block, machinery whose only job is moving a live Bitnami-era PVC through the one-time migration. Keeping it out avoids the footgun of a stray `recreateStatefulSet.enabled: true` deleting a StatefulSet; the trade-off is that a `claimName` rename now needs a manual STS delete. |
| `97950c6`, `b379b6d`, `f9d9ddb` | Deferred, not superseded. Their combined end state actually *disables* the S3 chart push (`echo "SKIP: helm s3 push disabled..."`) as a workaround for a corrupted bucket index. Not something to carry onto `main` silently; revisit if it resurfaces. |
| `aaac905` | No-op — the `release/*` Jenkinsfile trigger is already on `main` via #588/#589. |
| `518eac5` (#565) | Superseded within the source branch itself by `7c13757`. |

## Test plan

Run locally with `helm v4.1.4`:

- [x] All four charts lint clean.
- [x] etcd honours `replicaCount=3` and renders the PDB at >1 replica.
- [x] Scrape annotations appear with `podMonitor` off, and are suppressed with it on.
- [x] etcd PodMonitor renders only with `--api-versions monitoring.coreos.com/v1` — 0 without, 1 with.
- [x] The operator's `fix-helm-orphans` init container and all six `RELATED_IMAGE_*` env vars survive.
- [x] `konk` chart passes `replicaCount: 3` through to the Etcd CR.
- [x] All three dashboard JSONs parse; no selector in any of them has duplicate matchers on one label.
- [x] konk-service: no scheduling emitted by default, all 3 deployments still render, and setting `tolerations` propagates to all 3.
- [x] `claimName`: the `volumeClaimTemplate`, the container `volumeMount` and the `persistence.enabled=false` emptyDir all render `data-v2` consistently; overriding back to `data` still works; the konk chart sets no `claimName`, so clusters inherit the default through the Etcd CR.
- [x] CI green on this branch.

## Reviewer notes

1. **konk-service scheduling is now opt-in (`843ddc6`).** `main` shipped the
   `do-not-disrupt` toleration and stable-node-pool affinity *on by default* via #682.
   This PR makes them empty with a commented opt-in template, so konk-service pods lose
   that protection on next upgrade unless the KonkService CR or cluster values set them.
   Deliberate, and scoped to konk-service only — the `etcd` and `konk` charts keep their
   defaults, since etcd is the stateful workload the Karpenter consolidation incident
   (DEVOPS-47739) actually disrupted.
2. **The konk-operator PodMonitor may not render.** Its `monitoring.coreos.com/v1`
   capability gate is false under any *offline* `helm template` render (Argo CD's default
   mode included) and true when installing against a live cluster. Worth confirming which
   applies before relying on it for operator metrics — and do not pair it with
   `podAnnotations: null`, which would remove the annotation scrape path as well.
3. **Known limitation.** Three `konk-operator.json` panels still select on
   `app_kubernetes_io_instance="konk-operator"` literally, so they resolve only when the
   release is named `konk-operator`. `1b06f5d` templated this from `.Release.Name` and was
   reverted in `2438709`, so the hardcoded form is what merges here.
4. **`Recreate` narrows the two-operator window but does not close it structurally
   (`4c1bf77`).** Two gaps remain, both pre-existing and neither addressed here:
   `replicaCount > 1` would reinstate the race permanently, with no leader election to
   arbitrate — nothing sets it today (checked all 9 DC `konk-operator-values.yaml` files),
   but it is unguarded and would look like a reasonable HA change to make. And
   `deleteNonDeployedReleases` in `fix_helm_orphans_init.go` still deletes `pending-*`
   release secrets, which are indistinguishable from an upgrade in progress; narrowing it
   to `failed`/`unknown` is worthwhile defence in depth. (The *hook* variant,
   `fix_helm_orphans.go`, only patches annotations and never deletes — only the init
   variant deletes.)

`konk-service/README.md` is CI-generated (`readme` job → `make chart-readmes`) and is
regenerated on this branch automatically.


rsatal:
## Scope: what's included and what's deliberately excluded

This branch was cut fresh from `main` and cherry-picks from `release/upgrade-etcd`
(range `8b64bf7..release/upgrade-etcd`, 25 commits). Auditing that range against
`main` first, **9 of the 25 were already on `main`** and most of the rest turned out
to be superseded, redundant, or actively unsafe to replay. What's left is the subset
below.

### Included

| Commit | Source | What |
|---|---|---|
| `f448d9b` | #631 | Parameterizes the etcd `volumeClaimTemplate` name via `persistence.claimName` (default `data`). |
| `7190e58` | `9deaa9d` | etcd image `3.7.0` → `3.7.1`, fixing **GHSA-hrxh-6v49-42gf** (HIGH, grpc in the etcd/etcdctl/etcdutl binaries). `main` was still on `3.7.0`. |
| `b95ebb0` | #688 #689 #690 #691 | Grafana dashboards + metrics scraping — see below. |
| `843ddc6` | — | Makes konk-service pod scheduling preferences opt-in. |

**`7190e58` reconciliation notes.** Applied by hand rather than straight cherry-pick,
because the upstream commit was written against an older tree:
- Kept `main`'s parameterized `ETCD_IMG ?= gcr.io/etcd-development/etcd:$(ETCD_VERSION)`
  and bumped `ETCD_VERSION` to `v3.7.1`, instead of taking the hardcoded form.
- Chart version continues upward (`1.2.0` → `1.2.1`) instead of regressing to the
  upstream commit's `1.1.1`.

**`b95ebb0` (observability) detail.** The four dashboard commits all pile onto the same
three JSON files, so this takes their **final state per file** in one pass rather than
replaying four commits and resolving the same conflicts four times. Net result is
identical.
- `etcd.json` 12 → 17 panels, `konk-operator.json` 5 → 27 panels, and a new
  `konk-services.json` (38 panels) with its ConfigMap template.
- Verified the branch's dashboards are a **strict superset** of `main`'s — nothing is
  lost. The one apparently `main`-only panel, `Proposals Failed / Pending`, was *split*
  into separate Failed/Pending panels plus a new Committed Rate panel; both original
  metrics are retained.
- #684's Grafana 10 work survives: `dashboard.yaml` is byte-identical to `main`'s and
  both dashboards are all-`timeseries` with zero deprecated `graph` panels.
- konk-operator gains a `metrics` values block, a named `metrics` container port, and a
  PodMonitor for its own `/metrics`.
- etcd PodMonitor is now gated on the `monitoring.coreos.com/v1` capability, so enabling
  metrics can't fail the install on a cluster with no prometheus-operator.
- etcd emits the `prometheus.io` scrape annotations **only** when the PodMonitor is off,
  so collectors honouring both mechanisms stop scraping each pod twice.

`konk-operator/values.yaml` and `deployment.yaml` were edited surgically rather than
copied, because the upstream branch predates `main`'s `fix-helm-orphans` init container,
its `relatedImages` plumbing, and the move to the GHCR image repo.

### Excluded

| Commit(s) | Why |
|---|---|
| `f8540d7` | ⛔ **Would cause an outage.** Flips the etcd chart to read `.Values.statefulset.replicaCount` but doesn't touch `konk/values.yaml`, which still sends `replicaCount`. `main` is consistent on the flat key end-to-end and renders 3 replicas correctly. Applied alone this drops every konk-deployed etcd from 3 replicas to 1 — the exact failure `2606165`'s own commit body describes: *"causing the StatefulSet to scale from 3 to 1 on upgrade and leaving etcd-0 with stale 3-node cluster data that failed health checks."* |
| `2606165` (#683) | Fully redundant — self-described as a port of `main`'s #682, and `main` has the original plus everything since. PDB template, `pdb` values, etcd CPU/memory requests and the CI timeout are all already there (the timeout is already more generous). Taking it would also regress konk-service, which is further ahead on `main`. |
| `1de007e` (#635) | Superseded by `main`'s own `89ed1cf` (#637), the same fix re-landed 4 days later. The `helmAnnotations` helper and its ~33 template usages are already present. |
| `e13f6ed` | Superseded — `main` already moved to `cgr.dev/infoblox.com/etcd:3.7.0` via `b4dd9fa`, three weeks earlier. |
| `895772e`-adjacent hook work: `fd9ed6b` (#634), `372db4e` (#636) | Migration-only. Both touch nothing but `recreate-statefulset-hook.yaml` (plus its values block) — machinery whose only job is moving a live Bitnami-era PVC through the one-time migration. `main` only ever does fresh installs at the new chart version, so it has no job to do here, and a stray `recreateStatefulSet.enabled: true` would delete a StatefulSet. |
| `97950c6`, `b379b6d`, `f9d9ddb` | Deferred, not superseded. Their combined end state actually *disables* the S3 chart push (`echo "SKIP: helm s3 push disabled..."`) as a workaround for a corrupted bucket index. Not something to carry onto `main` silently; revisit if it resurfaces. |
| `aaac905` | No-op — the `release/*` Jenkinsfile trigger is already on `main` via #588/#589. |
| `518eac5` (#565) | Superseded within the source branch itself by `7c13757`. |

### Verification

All checks below were run locally with `helm v4.1.4` before pushing:
- All four charts lint clean.
- etcd honours `replicaCount=3` and renders the PDB at >1 replica.
- Scrape annotations appear with `podMonitor` off and are suppressed with it on.
- etcd PodMonitor renders only with `--api-versions monitoring.coreos.com/v1` — 0 without, 1 with.
- The operator's `fix-helm-orphans` init container and all six `RELATED_IMAGE_*` env vars survive.
- `konk` chart passes `replicaCount: 3` through to the Etcd CR.
- All three dashboard JSONs parse.
- konk-service: no scheduling emitted by default, all 3 deployments still render, and setting `tolerations` propagates to all 3.

### Two things a reviewer should know

1. **konk-service scheduling is now opt-in (`843ddc6`).** `main` shipped the
   `do-not-disrupt` toleration and stable-node-pool affinity *on by default* via #682.
   This PR makes them empty with a commented opt-in template, so konk-service pods lose
   that protection on next upgrade unless the KonkService CR or cluster values set them.
   The deployment templates wire both through, so opting in needs no chart change.
   Deliberate, and scoped to konk-service only — the `etcd` and `konk` charts keep their
   defaults, since etcd is the stateful workload the consolidation incident actually
   disrupted.
2. **The konk-operator PodMonitor may not render.** Its `monitoring.coreos.com/v1`
   capability gate is false under any *offline* `helm template` render (Argo CD's default
   mode included), and true when installing against a live cluster. Worth confirming
   which applies before relying on it for operator metrics — and don't pair it with
   `podAnnotations: null`, which would remove the annotation scrape path as well.

`konk-service/README.md` is CI-generated (`readme` job → `make chart-readmes`) and will
be regenerated on this branch automatically.

rsatal:
### Added: `strategy: Recreate` on the konk-operator Deployment (`4c1bf77`)

Adding a note here since this commit isn't part of the etcd port and reviewers will reasonably ask why it's in this PR.

**Why it's here.** The `End To End (true, v1.25.11, main)` leg was failing intermittently on this branch. It turned out not to be anything in the etcd work — it's a pre-existing race on `main` that this PR happens to trigger far more often than most.

**Root cause.** The operator Deployment had `replicas: 1` and no `strategy:` block, so Kubernetes defaults applied: `maxUnavailable: 25%` → floor → **0**, `maxSurge: 25%` → ceil → **1**. Since one pod must stay Available, the new pod is required to become Ready *before* the old one is removed — so **every** operator upgrade guaranteed a window with two operator pods running simultaneously.

Nothing arbitrates between them. The chart ships a leader-election `RoleBinding`, but the operator container is never passed `--leader-elect`, and the `Role` it binds to only exists in `config/rbac/`, not in this chart. So both instances reconcile the same `Konk` and `Etcd` CRs and drive Helm releases concurrently, corrupting each other's release history:

```
failed to update: secrets "sh.helm.release.v1.runner-konk-etcd.v3" not found
failed to install release: release: already exists
Error rolling back release ... rollback failed: release has no 1 version
```

Once the history is gone the operator can't converge, etcd never becomes healthy, and the konk apiserver never passes `readyz` — which is what tripped the `Test Konk` step.

**Why this PR surfaced it.** It's load-correlated, not random. #681 touched only the `etcd` chart, so a single release was in flight during the overlap — it passed 3/3. This branch changes the `konk`, `etcd` **and** `konk-service` charts, putting three releases in flight in the same window. It needed three attempts on an unchanged commit (fail, fail, pass) to go green.

**Secondary benefit.** The overlap also invalidated the `fix-helm-orphans` init container, which states it must "start with a clean slate on every operator restart". It deletes non-deployed release secrets, and `pending-upgrade` is indistinguishable from an upgrade the outgoing operator is running *right now*. `Recreate` makes that premise true rather than false by construction.

**Cost.** A few seconds with no operator running during an upgrade. Reconciliation resumes on startup and no state is lost — the right trade for a singleton that owns Helm releases.

**Alternatives considered.**
- *Enable `--leader-elect`* — wouldn't fix it. The init container runs *before* the operator container, so before any leader lock exists; it would still delete the outgoing leader's pending secrets. Also needs RBAC work for the missing `Role`.
- *Disable `initOrphanFix` in CI only* — mechanically easy, but the upgrade job would stop testing the real production upgrade path, and production has the same guaranteed overlap. That hides the defect rather than fixing it.
- *Narrow `deleteNonDeployedReleases` to skip `pending-*`* — good defence-in-depth and still worth doing, but it makes one symptom survivable instead of removing the cause.

**Verification.** `helm template` shows `strategy: Recreate` with no `maxSurge`/`maxUnavailable`/`RollingUpdate` keys emitted; the `fix-helm-orphans` init container and all six `RELATED_IMAGE_*` env vars are unaffected; all four charts lint clean.

vijaylakshmi-subramanian (approved):

